PR #31, opened 2021-01-07
FABGW-6 Enchance scenario tests for discovery

andrew-coleman:
Enhance the cucumber test framework to allow discovery testing.
Test network now comprises 5 peers across 3 orgs
Peers can be stopped and started to test different signature policy scenarios


pkg/server/protoutils.go
```diff
@@ -140,37 +143,32 @@ func createUnsignedTx(
...
-
-	_, err := rand.Read(key)
-	if err != nil {
-		return nil, errors.Wrap(err, "error getting random bytes")
-	}
-	return key, nil
-}
-
-func getChannelHeaderFromSignedProposal(signedProposal *peer.SignedProposal) (*common.ChannelHeader, error) {
+func getChannelAndChaincodeFromSignedProposal(signedProposal *peer.SignedProposal) (string, string, error) {
```
bestbeforetoday, line 146: The channel name is a top-level property of the ProposedTransaction so we don't need to deserialise and dig into the proposal to get it. I guess we should either:
1. Also move the chaincode name to the top-level of the ProposedTransaction, which means we can save a whole bunch of expensive deserialisation on the proposals within the Gateway; or
2. Just take the hit on deserialising the proposal and remove any ProposedTransaction properties that are duplicates of content in the proposal for simplicity.

But we could do that work outside of this PR, just as long as we actually do it.

scenario/fixtures/chaincode/golang/echo/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/hyperledger/fabric-gateway/scenario/echo/go
+
+go 1.15
+
+require (
+	github.com/hyperledger/fabric-contract-api-go v1.1.1
+	github.com/pkg/errors v0.9.1
```
bestbeforetoday, line 7: As per previous comments, I'm not sure it's a good idea to use this package when the core libraries give us similar functionality

java/src/test/java/scenario/ScenarioSteps.java
```diff
@@ -297,6 +310,34 @@ public ScenarioSteps() throws IOException {
...
+    }
+
+    private static void startAllPeers() throws InterruptedException {
+        runningPeers.forEach((peer, running) -> {
+            if (!running) {
+                try {
+                    exec("docker", "start", peer);
+                } catch (IOException e) {
+                    e.printStackTrace();
+                    return;
```
bestbeforetoday, line 332: If this fails the tests are broken so I think we'd be better throwing an Exception and failing fast instead of continuing on for the tests to behave unexpectedly later

java/src/test/java/scenario/ScenarioSteps.java
```diff
@@ -149,65 +164,59 @@ public ScenarioSteps() throws IOException {
...
-                            "peer0.org2.example.com:8051", "--tlsRootCertFiles",
+                    Collections.addAll(commitCommand, "docker", "exec", "org1_cli", "peer", "lifecycle", "chaincode", "commit",
+                            "--channelID", channelName,
+                            "--name", ccName,
+                            "--version", version,
+                            "--signature-policy", signaturePolicy,
+                            "--sequence", "1",
+                            "--waitForEvent",
+                            "--peerAddresses", "peer0.org1.example.com:7051",
+                            "--peerAddresses",
```
bestbeforetoday, line 217: Do we also need org3 here now we're a 3 org network?

scenario/node/lib/scenario_steps.js
```diff
@@ -244,13 +252,46 @@ When('I invoke the transaction', async () => {
...
+When('I invoke the transaction which I expect to fail', async () => {
+  const proposal = this.contract.newProposal(this.txn.name, this.txn.options);
+  try {
+    if (this.txn.type === 'evaluate') {
+      this.txn.result = await proposal.evaluate();
+    } else if (this.txn.type === 'submit') {
+      const transaction = await proposal.endorse();
+      this.txn.result = await transaction.submit();
+    } else {
+      //throw new Error(`Unknown transaction type: ${this.txn.type}`);
```
bestbeforetoday, line 264: The tests are broken if we get here so we should throw

bestbeforetoday: Actually, most of this function is a duplicate of the implementation of `When('I invoke the transaction')`. Perhaps we could factor out some common code

bestbeforetoday (changes requested):
I think there are a couple of things worth tweaking but the starting/stopping of peers in the tests is great. I like moving the expected fail to the transaction invocation too, so we can resubmit exactly the same transaction after restarting peers.

bestbeforetoday:
There is some overlap with #32. Your call on which one looks easier to get in first and rebase the other one on.

java/pom.xml
```diff
@@ -239,10 +239,11 @@
                     <show>public</show>
                     <doctitle>Hyperledger Fabric Gateway SDK for Java</doctitle>
                     <nohelp>true</nohelp>
-                    <failOnError>true</failOnError>
+                    <failOnError>false</failOnError>
```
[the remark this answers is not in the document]
andrew-coleman: oops, didn't mean to commit this!

java/src/test/java/scenario/ScenarioSteps.java
andrew-coleman: We could add it, but don't need to.  It just needs a majority of endorsers

scenario/node/lib/scenario_steps.js
andrew-coleman: > The tests are broken if we get here so we should throw

It will throw, but admittedly with a misleading message - I'll fix that.

pkg/server/protoutils.go
andrew-coleman: ok, this will all need revisiting anyway when we decide exactly what goes into the top-level message

bestbeforetoday (approved):

PR #60, opened 2021-03-24
Update npm prerelease version

jt-nti:
npm publish is failing with

“403 a package version that is forbidden by your security policy”

which may be due to the prerelease version string


jt-nti:
Looks like the 403 is due to an existing private package of the same name

PR #78, opened 2021-03-30
Gateway Java SDK sample

andrew-coleman:

### DIFF
--- a/samples/java/src/main/java/com/example/Sample.java
+++ b/samples/java/src/main/java/com/example/Sample.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.example;
+
+import java.io.FileReader;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.PrivateKey;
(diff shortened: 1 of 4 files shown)


samples/java/src/main/java/com/example/Sample.java
```diff
@@ -0,0 +1,59 @@
...
+import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
+import org.hyperledger.fabric.client.Contract;
+import org.hyperledger.fabric.client.Gateway;
+import org.hyperledger.fabric.client.Network;
+import org.hyperledger.fabric.client.identity.*;
+
+public class Sample
+{
+    private static final String mspID        = "Org1MSP";
+    private static final String cryptoPath   = "../../scenario/fixtures/crypto-material/crypto-config/peerOrganizations/org1.example.com";
```
jt-nti, line 23: It would be nice to avoid `/` path separators. E.g. using something like this instead...
```
private static final Path cryptoPath = Paths.get("..", "..", "scenario", "fixtures", "crypto-material", "crypto-config", "peerOrganizations", "org1.example.com");
```

jt-nti (approved):
Looks good. Just one suggestion for the paths.

jt-nti (approved):

PR #85, opened 2021-04-06
Add debug to scenario test

denyeart:
Add debug statements to scenario test.

Print CLI output even if an error is returned,
since the error often just includes an exit status,
while the output includes other useful debug information.


lesleyannjordan (approved):
LGTM

PR #86, opened 2021-04-06
Update scenario docker compose couchdbs

denyeart:
Update scenario docker compose so that each peer has its
own couchdb.

Also make misc docker-compose improvements.


lesleyannjordan (approved):
LGTM

PR #93, opened 2021-04-22
FABGW-13 Allow connection only with grpc channel

andrew-coleman:
Remove ability to connect to gateway peer using url/tls options.
Only support connections using client provided grpc channels.
This removes the problem of the slim SDKs having to provide the miriad of grpc options to the client, and encourages channel sharing across multiple connections.


bestbeforetoday (approved):
Much cleaner with all those connection options removed! 👍

PR #102, opened 2021-04-28
FABGW-10 Update samples to show async submit

andrew-coleman:
Add extra example code to the 3 SDK samples to show how to
submit multiple transactions without blocking the main thread, and later awaiting commit before querying the value.

Add running of samples to CI

### DIFF
--- a/samples/java/src/main/java/com/example/Sample.java
+++ b/samples/java/src/main/java/com/example/Sample.java
@@ -11,13 +11,19 @@ import java.nio.file.Paths;
 import java.security.PrivateKey;
 import java.security.cert.X509Certificate;
 import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
 import io.grpc.Channel;
 import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
 import org.hyperledger.fabric.client.Contract;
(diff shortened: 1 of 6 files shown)


samples/java/src/main/java/com/example/Sample.java
```diff
@@ -51,13 +57,37 @@ public static void main( String[] args ) throws Exception
...
+
+            // Submit multiple transactions asynchronously
+            List<Supplier<TransactionPackage.TxValidationCode>> commits = new ArrayList<>();
+            for (int i = 0; i < 5; i++) {
+                Transaction txn = contract.newProposal("put")
+                        .addArguments("key" + i, "value" + i)
+                        .build()
+                        .endorse();
+                System.out.println("Proposal result = " + new String(txn.getResult()));
+                Supplier<TransactionPackage.TxValidationCode> commit = txn.submitAsync();
```
bestbeforetoday, line 77: The `Supplier<TransactionPackage.TxValidationCode>` is equivalent to the `Commit` in Go and Node. It should probably be a `Commit` and not a plain `Supplier` but regardless... this code should be:
```
Transaction txn = contract.newProposal("put")
                        .addArguments("key" + i, "value" + i)
                        .build()
                        .endorse();
                Supplier<TransactionPackage.TxValidationCode> commit = txn.submitAsync();
                System.out.println("Proposal result = " + new String(txn.getResult()));
```
So the result is printed after submit to the orderer is successful to be consistent with the Go and Node samples.

bestbeforetoday (approved):

PR #111, opened 2021-04-29
FABGW-19: Use secured CommitStatus service

bestbeforetoday:
Also implements off-line signing flow for commit.


andrew-coleman (approved):
Looks good.  Once minor comment, but shouldn't hold us up.

PR #113, opened 2021-05-08
Fix the link of `Fabric Gateway RFC`.

yiwenlong:
Original link is not accessible.


bestbeforetoday:
I made a change to one of the gRPC services in Fabric, which is causing this PR build to fail. The corresponding client change is in #111 so this PR will need to rebased on top of that change once it is merged.

yiwenlong:
> I made a change to one of the gRPC services in Fabric, which is causing this PR build to fail. The corresponding client change is in #111 so this PR will need to rebased on top of that change once it is merged.

ok.

andrew-coleman:
The link is now working https://hyperledger.github.io/fabric-rfcs/text/0000-fabric-gateway.html
The breakage was caused by the renaming of branch `master` to `main`.
Thanks for pointing it out :)

PR #123, opened 2021-05-26
FABGW-20 SDK support for endorsing orgs

andrew-coleman:
Added ability to specify endorsing orgs to the 3 SDKs.
Added new scenario test and implementation in the 3 SDKs.

### DIFF
--- a/java/src/main/java/org/hyperledger/fabric/client/Proposal.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Proposal.java
@@ -74,6 +74,16 @@ public interface Proposal {
         Builder putTransient(String key, byte[] value);
 
         /**
+         * Specifies the set of organizations that will attempt to endorse the proposal.
+         * No other organizations' peers will be sent this proposal.
+         * This is usually used in conjunction with {@link #putTransient(String, byte[])} or
+         * {@link #putAllTransient(Map)} for private data scenarios.
+         * @param mspids The Member Services Provider IDs of the endorsing organizations.
+         * @return This builder.
+         */
+        Builder setEndorsingOrganizations(String... mspids);
+
+        /**
          * Build the proposal from the configuration state of this builder.
          * @return A proposal.
          */
--- a/java/src/main/java/org/hyperledger/fabric/client/ProposalImpl.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/ProposalImpl.java
@@ -66,6 +66,7 @@ final class ProposalImpl implements Proposal {
                 .setTransactionId(proposedTransaction.getTransactionId())
                 .setChannelId(channelName)
                 .setProposedTransaction(proposedTransaction.getProposal())
+                .addAllEndorsingOrganizations(proposedTransaction.getEndorsingOrganizationsList())
                 .build();
         EndorseResponse endorseResponse = client.endorse(endorseRequest);
 
--- a/node/src/transaction.test.ts
+++ b/node/src/transaction.test.ts
@@ -94,6 +94,12 @@ describe('Transaction', () => {
         expect(actual).toBe(expectedResult);
     });
 
+    it('sets endorsing orgs', async () => {
+        await contract.submit('TRANSACTION_NAME', { endorsingOrganizations: ['org1', 'org3']});
+        const actualOrgs = client.endorse.mock.calls[0][0].endorsing_organizations;
+        expect(actualOrgs).toStrictEqual(['org1', 'org3']);
+    });
+
     it('uses signer for submit', async () => {
         signer.mockResolvedValue(Buffer.from('MY_SIGNATURE'));
 
--- a/scenario/features/privatedata.feature
+++ b/scenario/features/privatedata.feature
@@ -1,0 +1,42 @@
+#
+# Copyright 2021 IBM All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+Feature: Private data
+    Background:
+        Given I have deployed a Fabric network
+        And I have created and joined all channels
+        And I deploy golang chaincode named basic at version 1.0.0 for all organizations on channel mychannel with endorsement policy OR("Org1MSP.member","Org2MSP.member","Org3MSP.member")
+        And I create a gateway for user User1 in MSP Org1MSP
+        And I connect the gateway to peer0.org1.example.com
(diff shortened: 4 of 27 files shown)


jt-nti (dismissed):

node/src/transaction.test.ts
```diff
@@ -94,6 +94,23 @@ describe('Transaction', () => {
...
+            actualOrgs = req.endorsing_organizations;
+            return Promise.resolve({
+                prepared_transaction: {
+                    payload: Buffer.from('PAYLOAD'),
+                },
+                result: {
+                    payload: Buffer.from(expectedResult),
+                },
+            });
+        })
```
bestbeforetoday, line 109: `client.endorse` is already a correctly configured mock function so you don't need to redefine it. After the call you can grab the request argument the mock captures with
```
const actualOrgs = client.endorse.mock.calls[0][0].endorsing_organizations;
```

scenario/features/privatedata.feature
```diff
@@ -0,0 +1,42 @@
...
+        Given I have deployed a Fabric network
+        And I have created and joined all channels
+        And I deploy golang chaincode named basic at version 1.0.0 for all organizations on channel mychannel with endorsement policy OR("Org1MSP.member","Org2MSP.member","Org3MSP.member")
+        And I create a gateway for user User1 in MSP Org1MSP
+        And I connect the gateway to peer0.org1.example.com
+        And I use the mychannel network
+        And I use the basic contract
+
+    Scenario: Submit to org1
+        When I prepare to submit an getPeerOrg transaction
```
bestbeforetoday, line 17: I hate to be that grammar lawyer but... "submit **a** getPeerOrg transaction"? 😉

scenario/fixtures/chaincode/golang/basic/main.go
```diff
@@ -77,3 +109,12 @@ func toStringValues(input map[string][]byte) map[string]string {
...
 }
+
+func contains(slice []string, entry string) bool {
+	for _, item := range slice {
+		if entry == item {
+			return true
+		}
+	}
+	return false
+}
```
bestbeforetoday, line 120: I don't see this used anywhere. Could it be omitted?

scenario/node/src/fabric.ts
```diff
@@ -209,8 +228,8 @@ export class Fabric {
             '--tlsRootCertFiles',
             '/etc/hyperledger/configtx/crypto-config/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/tls/ca.crt');
 
-        this.runningChaincodes[mangledName] = true;
-        await sleep(10000);
+        this.runningChaincodes[mangledName] = signaturePolicy;
+        await sleep(30000);
```
bestbeforetoday, line 232: For another day, but we really should find something better than this. Maybe polling the command to list running chaincodes until it appears?

bestbeforetoday (changes requested):
Generally looking good. A few code tweaks I think are worth addressing.

I'm not sure if it's a good idea to keep adding more things to the _basic_ chaincode. Maybe we should have a different chaincode for specific scenarios like private data? It might (or might not!) also reduce the work redeploying because of different endorsement policies. We can figure that out in another PR though.

java/src/main/java/org/hyperledger/fabric/client/Proposal.java
```diff
@@ -73,6 +73,16 @@
          */
         Builder putTransient(String key, byte[] value);
 
+        /**
+         * Specifies the set of organizations that will attempt to endorse the proposal.
+         * No other organizations' peers will be sent this proposal.
```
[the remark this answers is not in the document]
andrew-coleman: This describes how the gateway behaves at the moment.  It overrides the endorsement policy and ensures that private data won't leak to other orgs.  The requirements for state-based endorsement will be addressed in a future user story.

node/src/transaction.test.ts
andrew-coleman: Thanks - that was the magic I was looking for :)

scenario/features/privatedata.feature
andrew-coleman: quite right :)

scenario/fixtures/chaincode/golang/basic/main.go
andrew-coleman: Yes, it got left in there from an earlier experiment

scenario/node/src/fabric.ts
andrew-coleman: I wonder if these sleeps are carried over from the old chaincode lifecycle.  We are now specifying `--waitForEvent` on the commit so hopefully these sleeps aren't needed at all now.  I've taken them out and it works locally, so see what it does in the CI.

andrew-coleman: Oh well, that was a partial success.  Failed in Go - looks like the new endorsement policy hadn't been picked up in the first private-data test.  Adding short sleep back in. Not sure what the `--waitForEvent` is doing then.

java/src/main/java/org/hyperledger/fabric/client/Proposal.java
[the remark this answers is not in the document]
bestbeforetoday: 👍

scenario/node/src/fabric.ts
bestbeforetoday: My _guess_ is that the `--waitForEvent` is waiting for the config update to be committed, but at that point some of the behind-the-scenes deployment work may not yet be complete

java/src/main/java/org/hyperledger/fabric/client/ProposalImpl.java
```diff
@@ -66,6 +70,7 @@ public Transaction endorse() {
                 .setTransactionId(proposedTransaction.getTransactionId())
                 .setChannelId(channelName)
                 .setProposedTransaction(proposedTransaction.getProposal())
+                .addAllEndorsingOrganizations(Arrays.asList(endorsingOrgs))
```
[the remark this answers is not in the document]
andrew-coleman: It makes no difference as far as the behaviour of the gateway is concerned.  It treats empty array as unset.

bestbeforetoday (approved):

bestbeforetoday (approved):

PR #132, opened 2021-06-18
Add scenario tests for private collections and SBE

andrew-coleman:
Extend the privatedata.feature to test private data collections.
Add sbe.feature to test state-based endorsement.
Both sets of tests assert that setEndorsingOrgs function is working correctly.

The scenario test framework has been enhanced to allow multiple gateways to be defined and switched between within a scenario.


scenario/node/src/customworld.ts
```diff
@@ -95,31 +95,44 @@ function assertDefined<T>(value: T | null | undefined, property: string): T {
     return value;
 }
 
+interface GatewayContext {
+    gateway?: Gateway;
+    identity?: Identity;
```
bestbeforetoday, line 100: You always have an identity at the point you're creating the GatewayContext so this doesn't need to be optional. And you can save having the `getIdentity()` function to assert it exists.

bestbeforetoday (approved):
Those scenarios look much cleaner, and easier to understand what is being tested

PR #138, opened 2021-07-14
Add HSM Signing Support

davidkel:
Signed-off-by: D <d_kelsey@uk.ibm.com>

### DIFF
--- a/node/src/identity/hsmsigners.ts
+++ b/node/src/identity/hsmsigners.ts
@@ -1,0 +1,185 @@
+/*
+ * Copyright 2021 IBM All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as pkcs11js from 'pkcs11js';
+import * as elliptic from 'elliptic';
+import BN = require("bn.js");
+import ecSignature = require('elliptic/lib/elliptic/ec/signature');
+import { Signer } from './signer';
+
+type DER = string | number[];
+
+export interface HSMSignerOptions {
+    /**
+     * The label associated with the token for the slot
+     */
+    label: string;
+
+    /**
+     * The pin for the slot identified by the label
+     */
+    pin: string;
+
+    /**
+     * Identifier. The CKA_ID assigned to the HSM object
+     */
+    identifier: string | Buffer;
+
+    /**
+     * Optional user type for the HSM. If not specified it defaults to CKU_USER
+     */
+    userType?: number;
+}
+
+export interface HSMSigner {
+    signer: Signer;
+    close: () => void;
+}
+/**
+ * Create an HSM Signer factory. You should only ever call this once within your application
+ */
+export function newHSMSignerFactory(library: string): HSMSignerFactory {
+    if (!library || library.trim() === '') {
+        throw new Error('library must be provided');
+    }
+
+    return new HSMSignerFactory(library);
+}
+
+/**
+ * Factory to be able to create HSM Signers.
+ */
+export class HSMSignerFactory {
+
+    #pkcs11: pkcs11js.PKCS11;
+
+    constructor(library: string) {
+        this.#pkcs11 = new pkcs11js.PKCS11();
+        this.#pkcs11.load(library);
+        this.#pkcs11.C_Initialize();
+    }
+
+    /**
+     * disposes of the factory when it and HSM signers are not required anymore
+     */
+    public dispose(): void {
+        this.#pkcs11.C_Finalize();
+    }
+
+    /**
+     * Create a new HSM signing implementation based on provided HSM options.
+     *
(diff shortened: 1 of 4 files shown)


node/src/identity/hsmsigners.ts
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * This returns an object with 2 properties
+ * - signer which is the signer function
+ * - close which is a close function to close the signer when it's not required anymore
+ *
+ * @param hsmSignerOptions - The HSM signer options
+ * @returns an HSM Signer implementation
+ */
+export const newHSMSigner = (hsmSignerOptions: HSMSignerOptions): HSMSigner => {
+    if (!hsmSignerFactory) {
```
bestbeforetoday, line 58: This looks really dangerous. So the following code would end up with both signers unexpectedly using LIB_1?
```
const signer1 = newHSMSigner({ lib: 'LIB_1', ... });
const signer2 = newHSMSigner({ lib: LIB_2', ... });
```

It seems like we should either:
1. Create a new HSMSignerFactory each time; or
2. If creating multiple HSMSigner instances from the same factory is important (i.e. the PKCS11 load/initialize must only happen once), expose the HSMSignerFactory for the client to create and use to create HSM signers as required.

node/src/identity/hsmsigners.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    public newSigner(label: string, pin: string, userType: number = pkcs11js.CKU_USER, keySize: number, identifier: string | Buffer): {signer: Signer, close: () => void} {
+        const pkcs11 = this.#pkcs11;
+        const slot = this.findSlotForLabel(label);
+        const session = pkcs11.C_OpenSession(slot, pkcs11js.CKF_SERIAL_SESSION);
+        pkcs11.C_Login(session, userType, pin);
+        const privateKeyHandle = this.findObjectInHSM(session, pkcs11js.CKO_PRIVATE_KEY, identifier);
+
+        const curveName = `secp${keySize}r1`;
```
bestbeforetoday, line 101: This is only used for error messages. Is it worth losing this and just using the actual curve name we ask for (`p${keySize}`) instead?

node/src/identity/hsmsigners.ts
```diff
@@ -0,0 +1,186 @@
...
+            const signatureInput: elliptic.SignatureInput = {
+                r,
+                s
+            }
+
+            const der = new ecSignature(signatureInput).toDER() as DER; // eslint-disable-line
+            return Promise.resolve(Buffer.from(der));
+        }
+
+        return {signer, close};
```
bestbeforetoday, line 139: It might make the code less busy to pull this out to its own (unexported) class, but whatever you prefer is fine.

node/src/identity/hsmsigners.ts
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright 2021 IBM All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as pkcs11js from 'pkcs11js';
+import * as elliptic from 'elliptic';
+import BN = require("bn.js");
+const ecSignature = require('elliptic/lib/elliptic/ec/signature.js');  // eslint-disable-line
```
bestbeforetoday, line 10: Is there a public API that can be used rather than poking at internals? The SDKs have been broken before by using internals of dependencies that change when we pull in newer versions.

You might be able to lose the `.js` on the end of the required module, and generally prefer `import ... from ...` instead of `require`.

node/src/identity/hsmsigners.ts
```diff
@@ -0,0 +1,186 @@
...
+            pkcs11.C_SignInit(session, { mechanism: pkcs11js.CKM_ECDSA }, privateKeyHandle);
+            const sig = pkcs11.C_Sign(session, Buffer.from(digest), Buffer.alloc(keySize));
+
+            const r = new BN(sig.slice(0, sig.length / 2).toString('hex'), 16);
+            let s = new BN(sig.slice(sig.length / 2).toString('hex'), 16);
+            const halfOrder = ecdsaCurve.n?.shrn(1);
+
+            if (!halfOrder) {
+                throw new Error(`Can not find the half order needed to calculate "s" value for immalleable signatures. Unsupported curve name ${curveName}`);
+            }
```
bestbeforetoday, line 120: We know the curve before we do any signing. Could we obtain the half order for the curve up front and error during creation of the signer rather than during signing?

bestbeforetoday (changes requested):
The handling of PKCS11 library looks like a big gotcha for users so I think that really needs fixing. Several other more minor tweaks and suggestions to consider.

bestbeforetoday (approved):

PR #145, opened 2021-07-26
FABGW-25 Scenario tests for collections & SBE

andrew-coleman:
Added extra scenarios to the privatedata and sbe feature files to test the new endorser selection logic in the gateway

### DIFF
--- a/scenario/features/sbe.feature
+++ b/scenario/features/sbe.feature
@@ -35,3 +35,40 @@ Feature: State-based endorsement
         And I set the transaction arguments to ["key-001"]
         And I invoke the transaction
         Then the response should be "value-001"
+
+    Scenario: Set SBE policy for Org1 AND Org3; state change successfully when not specifying endorsing orgs
+        When I prepare to submit a SetStateEndorsers transaction
+        And I set the transaction arguments to ["key-001", "Org1MSP", "Org3MSP"]
+        And I invoke the transaction
+        And I prepare to submit a ChangeState transaction
+        And I set the transaction arguments to ["key-001", "value-Org1-Org3"]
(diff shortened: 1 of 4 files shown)


scenario/features/sbe.feature
```diff
@@ -35,3 +35,40 @@ Feature: State-based endorsement
...
+    Scenario: Set SBE policy for Org1 AND Org3; state change successfully when not specifying endorsing orgs
+        When I prepare to submit a SetStateEndorsers transaction
+        And I set the transaction arguments to ["key-001", "Org1MSP", "Org3MSP"]
+        And I invoke the transaction
+        And I prepare to submit a ChangeState transaction
+        And I set the transaction arguments to ["key-001", "value-Org1-Org3"]
+        And I invoke the transaction
+        And I prepare to evaluate a GetState transaction
+        And I set the transaction arguments to ["key-001"]
+        And I invoke the transaction
```
denyeart, line 48: Shouldn't we use a verb like "evaluate" or "query" here instead of "invoke"?

denyeart (approved):
LGTM.
Just one minor comment but it is about the already existing feature test rather than this PR, so I'll go ahead and merge.

scenario/features/sbe.feature
andrew-coleman: It's a neutral term because to evaluate a transaction we would use:

And I prepare to **evaluate** a GetState transaction
And I set the transaction arguments to ["key-001"]
And I invoke the transaction

To submit a transaction:

And I prepare to **submit** a GetState transaction
And I set the transaction arguments to ["key-001"]
And I invoke the transaction

PR #151, opened 2021-07-29
Update repository ID in Java pom.xml

jt-nti:
Java publish is broken due to a 401 error

Attempting to fix by changing the id to match the maven service connection ID…

“You should set the repositories in your project's pom.xml to have the same <id> as the name specified in the task for Maven to be able to correctly authenticate the task.”

https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/maven-authenticate?view=azure-devops

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -42,7 +42,7 @@
 
     <repositories>
         <repository>
-            <id>nightly-repo</id>
+            <id>maven-artifactory</id>
             <url>https://hyperledger-fabric.jfrog.io/artifactory/fabric-maven</url>
             <releases>
                 <enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


andrew-coleman (approved):

java/pom.xml
```diff
@@ -323,7 +323,7 @@
 
     <distributionManagement>
         <repository>
-            <id>fabric-maven</id>
+            <id>maven-artifactory</id>
```
denyeart, line 326: When I generate the settings from our jfrog artifactory, it gives me:

```
<distributionManagement>
    <repository>
        <id>central</id>
        <name>hyperledger-artifactory-primary-0-releases</name>
        <url>https://hyperledger-fabric.jfrog.io/artifactory/fabric-maven</url>
    </repository>
    <snapshotRepository>
        <id>snapshots</id>
        <name>hyperledger-artifactory-primary-0-snapshots</name>
        <url>https://hyperledger-fabric.jfrog.io/artifactory/fabric-maven</url>
    </snapshotRepository>
</distributionManagement>
```

jt-nti: The url matches, so that's good. Hopefully the id in the settings.xml that azure generates and the pom will now match and it'll start working again🤞

PR #156, opened 2021-08-05
Example of reading ids from JSON files

mbwhite:
Two main types of JSON file that contain ids. Example of reading those and then prcoessing them for use with the gateway

Signed-off-by: Matthew B White <whitemat@uk.ibm.com>

### DIFF
--- a/samples/java/src/main/java/com/example/JsonIdAdapter.java
+++ b/samples/java/src/main/java/com/example/JsonIdAdapter.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.example;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.security.InvalidKeyException;
+import java.security.PrivateKey;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+import java.util.logging.Logger;
+import javax.json.Json;
+import javax.json.JsonObject;
+
+import org.hyperledger.fabric.client.identity.Identities;
+import org.hyperledger.fabric.client.identity.Identity;
+import org.hyperledger.fabric.client.identity.Signer;
+import org.hyperledger.fabric.client.identity.Signers;
+import org.hyperledger.fabric.client.identity.X509Identity;
+
+/**
+ * This class can be used to map identities in a variety of JSON formats to the
+ * Identity and Signers required for the gateway. For example if you have an
+ * application wallet, or have exported IDs from SaaS
+ * 
+ * ``` 
+ * // setup the loading of the identities JsonIdAdapter idWallet = new
+ * JsonIdAdapter(Paths.get(walletDir)); Identity id =
+ * idWallet.getIdentity(userid); Signer signer = idWallet.getSigner(userid);
+ * Gateway.Builder builder =
+ * Gateway.newInstance().identity(id).signer(signer).connection(grpcChannel);
+ * 
+ * ```
+ * 
+ * Though they are JSON files, typically they files will have the .id extension.
+ * Therefore if no extension is provided `.id` is added
+ */
+public class JsonIdAdapter {
+    private static final Logger LOGGER = Logger.getLogger(JsonIdAdapter.class.getName());
+    public static final String JSON_VERSION = "version";
+    public static final String JSON_TYPE = "type";
+    public static final String JSON_MSP_ID = "mspId";
+
+    private static final String JSON_CREDENTIALS = "credentials";
+    private static final String JSON_CERTIFICATE = "certificate";
+    private static final String JSON_PRIVATE_KEY = "privateKey";
+    private static final String DATA_FILE_EXTENTION = ".id";
+
+    private final Path idFilesDir;
+
+    /** 
+     * @param idFilesDir
+     */
+    public JsonIdAdapter(final Path idFilesDir) {
+        this.idFilesDir = idFilesDir;
+
+        if (!Files.exists(idFilesDir)) {
+            throw new RuntimeException("Directory " + idFilesDir + " does not exist");
+        }
+
+        LOGGER.info("Reading for " + idFilesDir);
+
+    }
+
+    private Path getPathForLabel(final String label) {
+        return idFilesDir.resolve(label + DATA_FILE_EXTENTION);
+    }
+
+    public Identity getIdentity(String id) throws CertificateException, IOException {
+        JsonObject json = readFile(id);
+        String mspId = json.getString(JSON_MSP_ID);
+        JsonObject credentials = json.getJsonObject(JSON_CREDENTIALS);
+        String certificatePem = credentials.getString(JSON_CERTIFICATE);
+
+        X509Certificate certificate = Identities.readX509Certificate(new StringReader(certificatePem));
+        return new X509Identity(mspId, certificate);
+    }
+
+    public Signer getSigner(String id) throws InvalidKeyException, IOException {
+        JsonObject json = readFile(id);
+        JsonObject credentials = json.getJsonObject(JSON_CREDENTIALS);
+        String privateKeyPem = credentials.getString(JSON_PRIVATE_KEY);
+
+        PrivateKey privateKey = Identities.readPrivateKey(new StringReader(privateKeyPem));
+
+        return Signers.newPrivateKeySigner(privateKey);
+    }
+
+    private JsonObject readFile(final String label) throws IOException {
+        InputStream identityData = Files.newInputStream(getPathForLabel(label));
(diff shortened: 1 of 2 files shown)


samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    private readIDFile(idFile: string): JSONID {
+        let idJsonFile = path.resolve(path.join(this.idFilesDir, idFile));
+
+        // check if there's no extension probably means it's a waller id file
+        if (path.extname(idJsonFile)===""){
+            idJsonFile = `${idJsonFile}.id`
+        }
+
+        if (!fs.existsSync(idJsonFile)) {
```
bestbeforetoday, line 62: It might be better to just read the file and complain if the read fails as the state may change between existence check and reading.

samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+            throw new Error(`Can't locate the id file [${idJsonFile}]`);
+        }
+        let id: JSONID;
+        let json = JSON.parse(fs.readFileSync(idJsonFile, 'utf-8'));
```
bestbeforetoday, line 66: I prefer `await fs.promises.readFile(idJsonFile)` to sync calls as they block the Node event loop, whereas the promise implementation is handled entirely asynchronously on another thread.

mbwhite: np, can use the async variants

samples/node/src/jsonid-adapter.ts
mbwhite: fair point, will check it doesn't give a dreadful error though.

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+ * application wallet, or have exported IDs from SaaS
+ * 
+ * ``` 
+ * // setup the loading of the identities JsonIdAdapter idWallet = new
+ * JsonIdAdapter(Paths.get(walletDir)); Identity id =
+ * idWallet.getIdentity(userid); Signer signer = idWallet.getSigner(userid);
+ * Gateway.Builder builder =
+ * Gateway.newInstance().identity(id).signer(signer).connection(grpcChannel);
+ * 
+ * ```
```
bestbeforetoday, line 41: JavaDoc doesn't do markdown do you probably want:
```
<pre>{@code
// Code example here
}</pre>
```
And perhaps format so lines don't mix comments with trailing code, or contain multiple commands.

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+        X509Certificate certificate = Identities.readX509Certificate(new StringReader(certificatePem));
+        return new X509Identity(mspId, certificate);
+    }
+
+    public Signer getSigner(String id) throws InvalidKeyException, IOException {
+        JsonObject json = readFile(id);
+        JsonObject credentials = json.getJsonObject(JSON_CREDENTIALS);
+        String privateKeyPem = credentials.getString(JSON_PRIVATE_KEY);
+
+        PrivateKey privateKey = Identities.readPrivateKey(new StringReader(privateKeyPem));
```
bestbeforetoday, line 92: `Identities.readPrivateKey()` has overloads for both `Reader` and `String` parameters so you can just use `Identities.readPrivateKey(privateKeyPem)`

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+        return idFilesDir.resolve(label + DATA_FILE_EXTENTION);
+    }
+
+    public Identity getIdentity(String id) throws CertificateException, IOException {
+        JsonObject json = readFile(id);
+        String mspId = json.getString(JSON_MSP_ID);
+        JsonObject credentials = json.getJsonObject(JSON_CREDENTIALS);
+        String certificatePem = credentials.getString(JSON_CERTIFICATE);
+
+        X509Certificate certificate = Identities.readX509Certificate(new StringReader(certificatePem));
```
bestbeforetoday, line 83: `Identities.readX509Certificate()` has overloads for both `Reader` and `String` parameters so you can just use `Identities.readX509Certificate(certificatePem)`

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+    private static final String JSON_CREDENTIALS = "credentials";
+    private static final String JSON_CERTIFICATE = "certificate";
+    private static final String JSON_PRIVATE_KEY = "privateKey";
+    private static final String DATA_FILE_EXTENTION = ".id";
+
+    private final Path idFilesDir;
+
+    /** 
+     * @param idFilesDir
+     */
```
bestbeforetoday, line 61: Having no JavaDoc comment would be better than this.

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+import org.hyperledger.fabric.client.identity.Identities;
+import org.hyperledger.fabric.client.identity.Identity;
+import org.hyperledger.fabric.client.identity.Signer;
+import org.hyperledger.fabric.client.identity.Signers;
+import org.hyperledger.fabric.client.identity.X509Identity;
+
+/**
+ * This class can be used to map identities in a variety of JSON formats to the
+ * Identity and Signers required for the gateway. For example if you have an
+ * application wallet, or have exported IDs from SaaS
```
bestbeforetoday, line 32: It looks like this implementation is only handling v2 wallets

samples/java/src/main/java/com/example/JsonIdAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+ * // setup the loading of the identities JsonIdAdapter idWallet = new
+ * JsonIdAdapter(Paths.get(walletDir)); Identity id =
+ * idWallet.getIdentity(userid); Signer signer = idWallet.getSigner(userid);
+ * Gateway.Builder builder =
+ * Gateway.newInstance().identity(id).signer(signer).connection(grpcChannel);
+ * 
+ * ```
+ * 
+ * Though they are JSON files, typically they files will have the .id extension.
+ * Therefore if no extension is provided `.id` is added
```
bestbeforetoday, line 44: It looks like this implementation always adds the `.id` extension

samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+import { Identity, Signer, signers } from 'fabric-gateway';
+import fs from 'fs/promises';
+import path from 'path';
+import crypto from 'crypto';
```
bestbeforetoday, line 11: Sometimes I see TypeScript flag up errors for this import style, either depending on the tsconfig or the type of module being imported. I find the `import { promises as fs } from 'fs'` style typically causes fewer issues, but as long as it's building and working, whatever format you prefer is fine by me.

samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/** Internal interface used to describe all the possible components
```
bestbeforetoday, line 13: Only if you're making other changes, fixing this formatting might be good

samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+        return id;
+    }
+
+    /**
+     * 
+     * @param idFile the name of the identity to load (if no extension is provided `.id` is added)
+     * @returns Identity to use with the GatewayBuilder
+     */
```
bestbeforetoday, line 102: If you're including method doc then a description would be good

samples/node/src/jsonid-adapter.ts
```diff
@@ -0,0 +1,126 @@
...
+        };
+
+        return identity;
+    }
+
+    /**
+     * 
+     * @param idFile the name of the identity to load (if no extension is provided `.id` is added)
+     * @returns Signer to use with the GatewayBuilder
+     */
```
bestbeforetoday, line 118: If you're including method doc then a description would be good

samples/java/src/main/java/com/example/JsonIdAdapter.java
bestbeforetoday: For the example it may be more natural to create a Gateway instance rather than the Builder, which would match up with what you have for the Node sample:
```
Gateway gateway = Gateway.newInstance()
        .identity(id)
        .signer(signer)
        .connection(grpcChannel)
        .connect();
```

bestbeforetoday (changes requested):
Lots of comments but all just minor tweaks / polish. Generally looks good to me.

PR #162, opened 2021-08-12
Update gateway sample readme with additional information

denyeart:
Clean up instructions for v2.4.0-beta.
Add a description of the scenarios demonstrated in the sample client applications.
Minor cleanups to the scenario fixtures.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -3,29 +3,89 @@
 The samples in this repo show how to create client applications that invoke transactions using the new embedded Gateway
 in Fabric.
 
-The samples will only run against the latest version of Fabric - v2.4.0-alpha.  The easiest way of setting up a gateway
-enabled Fabric network is to use the scenario test framework that is part of this `fabric-gateway` repository using the
-following command:
+# Setup
+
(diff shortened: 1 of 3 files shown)


davidkel (commented):
There is also an hsm-samples/README.md file which also references 2.4.0-alpha don't know if you want to include that in this PR as well ?

jt-nti:
It would also be good to update [the doc](https://hyperledger.github.io/fabric-gateway/) from Tech Preview to Beta, with what's new etc. E.g. it still says "Currently the Gateway component in the Fabric Peer is disabled by default" which has been switched round I think.

bestbeforetoday:
> It would also be good to update [the doc](https://hyperledger.github.io/fabric-gateway/) from Tech Preview to Beta, with what's new etc.

That would definitely be good. It may be worth noting that limited (experimental) chaincode event listening capability is included, which does not yet support event replay.

bestbeforetoday (approved):
These changes all look good to me. Additional comments can be addressed in a follow-up PR.

PR #164, opened 2021-08-13
Support HSM Scenario tests for Go Signer

davidkel:
Signed-off-by: D <d_kelsey@uk.ibm.com>


bestbeforetoday:
@davidkel before you go too far with this one, I'm slightly sceptical of the value of off-line signing tests for HSM. The off-line signing tests just ensure that the off-line signing flow works for the client SDK and it really makes no difference what signing implementation is used. Either they do use the signing implementation to correctly generate signatures and include them in messages sent to Fabric, or they don't. Rather than add off-line signing tests for the Go HSM signer, I'd be more inclined to remove the off-line signing tests for the Node HSM signer to reduce the amount of code to be maintained and minimise test run times.

The straightforward HSM signing tests are definitely necessary to ensure that signatures generated by HSM signers in Fabric Gateway clients can be used with Fabric. It's just the off-line signing variations that I don't think are necessary.

bestbeforetoday (approved):

PR #180, opened 2021-08-30
Update Go to 1.16.7

denyeart:
Update Go to 1.16.7.
Also run "go mod tidy".


bestbeforetoday:
Could we also update all the `go.mod` files?

denyeart:
> Could we also update all the `go.mod` files?

Yes, I'll do.

bestbeforetoday:
The `go.mod` file still contains `go 1.15`. Would it be better for it to match the go version we’re specifying elsewhere? Same for the test chaincode at `scenario/fixtures/chaincode/go/basic/go.mod`

bestbeforetoday (approved):

PR #214, opened 2021-09-24
Refactor commit status API to improve usability

bestbeforetoday:
Minimise the number of potential error points that application code need to handle. Also minimise mutability within implementation.

Updated samples to reflect API change, which required samples to use the current development code.


java/src/main/java/org/hyperledger/fabric/client/Status.java
```diff
@@ -0,0 +1,61 @@
...
+     * Get the transaction ID.
+     * @return A transaction ID.
+     */
+    public String getTransactionId() {
+        return transactionId;
+    }
+
+    /**
+     * Get the block number in which the transaction committed. If the transaction has not yet committed, this method
+     * blocks until the commit occurs.
```
andrew-coleman, line 37: I guess it blocks on the getStatus() call now rather than here?

java/src/main/java/org/hyperledger/fabric/client/Status.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return A block number.
+     */
+    public long getBlockNumber() {
+        return blockNumber;
+    }
+
+    /**
+     * Get the committed transaction status code. If the transaction has not yet committed, this method blocks until
+     * the commit occurs.
```
andrew-coleman, line 46: Ditto

java/src/main/java/org/hyperledger/fabric/client/Status.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return Transaction status code..
+     */
+    public TxValidationCode getCode() {
+        return code;
+    };
+
+    /**
+     * Check whether the transaction committed successfully. If the transaction has not yet committed, this method
+     * blocks until the commit occurs.
```
andrew-coleman, line 55: and here

andrew-coleman (commented):
Just comments on the comments :)

java/src/main/java/org/hyperledger/fabric/client/Status.java
bestbeforetoday: Yes, you're right. That needs fixing

andrew-coleman (approved):

Issue #302, opened 2021-11-06
Specific client errors for each part of transaction flow


bestbeforetoday:
**As an** application developer
**I want** specific errors to be raised from each remote invocation in the transaction flow
**So that** I can programmatically identify the point in the flow that failed when using higher level functions and react appropriately

Specific errors for:

- [ ] Endorse
- [ ] Submit
- [ ] CommitStatus
- [ ] Unsuccessful commit status

These errors must include and make accessible programmatically:

- [ ] Transaction ID
- [ ] gRPC status
- [ ] gRPC error details

bestbeforetoday:
Relates to #178

PR #304, opened 2021-11-09
More specific errors in Node API

bestbeforetoday:
Add EndorseError, SubmitError and CommitStatusError to better differentiate the point in the submit flow where a failure occurs.

Contributes to #302


andrew-coleman (approved):
We might need to add another sample for this, but it can be done in a separate PR

PR #327, opened 2021-11-25
Initial release process documentation

jt-nti:

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,16 +2,62 @@
 
 The following artifacts are created as a result of releasing the Hyperledger Fabric Gateway SDKs:
 
-- TBC
+- npm modules
+    - [@hyperledger/fabric-gateway](https://www.npmjs.com/package/@hyperledger/fabric-gateway)
+- Java libraries
+    - [fabric-gateway](https://search.maven.org/artifact/org.hyperledger.fabric/fabric-gateway)
 
 ## Before releasing
 
-TBC
+The following tasks are required before releasing:
+
+- Update version numbers if required (see below for details)
+- Update test, sample, and docs files to match the new version if it was updated
 
 ## Create release
 
-TBC
+Creating a GitHub release on the [releases page](https://github.com/hyperledger/fabric-gateway/releases) will trigger the build to publish the new release.
+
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -2,16 +2,62 @@
 
 The following artifacts are created as a result of releasing the Hyperledger Fabric Gateway SDKs:
 
-- TBC
+- npm modules
+    - [@hyperledger/fabric-gateway](https://www.npmjs.com/package/@hyperledger/fabric-gateway)
+- Java libraries
+    - [fabric-gateway](https://search.maven.org/search?q=a:fabric-gateway)
```
bestbeforetoday, line 8: I think link might work better: https://search.maven.org/artifact/org.hyperledger.fabric/fabric-gateway

RELEASING.md
```diff
@@ -2,16 +2,62 @@
...
+
+The Hyperledger Fabric Gateway SDKs follow the [Go module version numbering system](https://go.dev/doc/modules/version-numbers)
+
+## Updating version numbers
+
+The following files need to be modified when updating the version number, and these are checked by the build process to ensure they match a tagged release:
+
+- The `GATEWAY_VERSION` variable in `ci/azure-pipelines.yml`
+- The `version` element in `java/pom.xml`
+- The `version` property in `node/package.json`
```
bestbeforetoday, line 55: If the samples are staying in this repo then, for Node and Java, probably we should update both their versions and fabric-gateway dependency versions

jt-nti: > Update test, sample, and docs files to match the new version

I was admittedly/deliberately a bit vague about the sample versions but those sort of things tend to change over time, whereas the versions here are a bit more fundamental

If the samples are staying (are they?), I'm not sure we should be updating the version of the samples (that doesn't happen in fabric-samples afaik), and I think it's only the Java dependency that needs updating and it turns out that the build breaks if you miss it :)

bestbeforetoday: I was thinking the sample version matching up to a fabric-gateway version would help to identify issues with mismatched version not working due to API changes. But I guess you're right, the samples are checked in with the code they match and you shouldn't really be using them as-is outside of that context. And I'm not convinced keeping the samples here is the best idea either.

bestbeforetoday (approved):

PR #379, opened 2022-01-31
Set correct channel name when getting contract

ppoffice:
Signed-off-by: Ruipeng Zhang &lt;ppoffice@users.noreply.github.com&gt;


andrew-coleman (approved):
Good spot - thanks

PR #407, opened 2022-03-15
Client API changes for checkpointer golang

sapthasurendran:


pkg/checkpoint/checkpointers.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package checkpoint
+
+type Checkpointer interface {
+    Checkpoint  (func() ( uint64, bool),  ...string)
+    GetBlockNumber() *BlockNumber
+    GetTransactionID() string
+}
+type CheckpointData struct {
```
bestbeforetoday, line 14: This type shouldn't be needed

bestbeforetoday, line 14: This type shouldn't be needed

pkg/client/eventsbuilder.go
```diff
@@ -44,3 +47,25 @@ func WithStartBlock(blockNumber uint64) eventOption {
 		return nil
 	}
 }
+
+// WithCheckpointer reads events starting at the specified start block and after the passed transaction ID .
+func WithCheckpointer(checkpointer *checkpoint.CheckpointData) eventOption {
```
bestbeforetoday, line 52:
```suggestion
// WithCheckpointer reads events starting immediately after the position recorded by the checkpointer.
func WithCheckpointer(checkpointer *checkpoint.Checkpointer) eventOption {
```

bestbeforetoday, line 52:
```suggestion
// WithCheckpointer reads events starting immediately after the position recorded by the checkpointer.
func WithCheckpointer(checkpointer *checkpoint.Checkpointer) eventOption {
```

bestbeforetoday (changes requested):
We are going to need to be careful in defining the checkpointer behaviour and API. I think we expect the checkpointer's block number to be the block at which we should start eventing. We expect any recorded transaction ID to be the last transaction ID successfully processed within that block, so eventing should resume immediately after that transaction.

This should work as expected for chaincode event checkpointing where the application is checkpointing only the current block number and last processed transaction ID. We need to be careful that it also works for block eventing, where the application is checkpointing only the last successfully processed block number. In this case we expect to resume from the following block, not to resume from the block we last processed as this would be duplication.

Perhaps the behaviour should be that if you checkpoint only a block number (N), it sets the checkpointer's block number to the next block (N + 1), whereas checkpointing a transaction uses exactly that block number as the checkpointer's block number.

I have a PR in flight that changes the Gateway service behaviour to automatically locate the correct start block for a given transaction ID, [here](https://github.com/hyperledger/fabric/pull/3289). We could take advantage of that behaviour and have the checkpointer not worry about the block number when checkpointing a transaction. Perhaps the behaviour could be:
- when checkpointing a block (supplying only the block number, N), the checkpointer sets its block number to the next block (N + 1) and unsets the transaction ID.
- when checkpointing a transaction (supplying only a transaction ID), the checkpointer unsets the block number and sets the transaction ID.

Assuming this behaviour works, we could even go further and have a separate block and transaction (chaincode event) checkpointer. I'm not sure whether that is helpful or more confusing for application developers.

bestbeforetoday:
Pulling on that thread of having checkpoint of a block (not a transaction) increment to the next block as that's now the point to resume from... perhaps this means that you can never be in a state after checkpointing anything where both the block number is zero and there is no transaction ID. So the "nil" state of a checkpointer would could be blockNumber = 0 and transactionID = "". As soon as you checkpoint, you either have a non-empty transactionID (along with the zero block number) or a non-zero block number. That would remove the need for storing an additional piece of information to say whether the block number had been set and the checkpointer interface could become just:
```go
type Checkpointer interface {
    BlockNumber() uint64
    TransactionID() string
}
```
What do you think?

bestbeforetoday:
And... if for Go we will have a `CheckpointBlock(blockNumber uint64)` and `CheckpointTransaction(blockNumber uint64, transactionID string)`, perhaps we should take the same approach in both Node and Java for consistency, rather than using method overloading of a single `checkpoint()` method. What do you think?

pkg/client/eventsbuilder.go
```diff
@@ -31,6 +33,12 @@ func (builder *eventsBuilder) getStartPosition() *orderer.SeekPosition {
 
 type eventOption = func(builder *eventsBuilder) error
 
+// Inferace used while checkpointing events
```
bestbeforetoday, line 36: We can fix this later when we check all the docs, but godoc comments must start with the name of the thing they are commenting. So The first word of this comment must be **Checkpointer**. See https://go.dev/blog/godoc

pkg/client/inmemorycheckpointer_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+package client
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestInMemoryCheckpointer(t *testing.T) {
```
bestbeforetoday, line 15: This should include unit test for `CheckpointChaincodeEvent()` too, but can add that later

bestbeforetoday (approved):

PR #414, opened 2022-03-23
file checkpointer for go

sapthasurendran:


pkg/client/checkpointer_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+type Checkpoint interface {
+	checkpoint
+	Checkpointer
+}
+
+type testCase struct {
+	description     string
+	after           func()
+	newCheckpointer func() Checkpoint
```
bestbeforetoday, line 32: Since this can error for a file checkpointer, perhaps `newCheckpointer` should return an error, which would always be `nil` for the in-memory checkpointer?
```suggestion
	newCheckpointer func() (Checkpoint, error)
```

Or if you never want to test error conditions, you could pass the Go test in to allow this to fail the test on error rather than ignore it:
```suggestion
	newCheckpointer func(t *testing.T) Checkpoint
```

pkg/client/checkpointer_test.go
```diff
@@ -0,0 +1,127 @@
...
+				fileCheckpointer, err := NewFileCheckpointer(path)
+				require.NoError(t, err)
+				return fileCheckpointer, err
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+
+		t.Run(tc.description+" :Initializes default checkpointer state when no checkpointer already exist", func(t *testing.T) {
```
bestbeforetoday, line 78: Rather than adding the testCase description to every test, it might be simpler to wrap them all in an enclosing test run, so all your tests for a specific checkpointer are grouped together
```suggestion
	for _, tc := range testCases {
		t.Run(tc.description, func(t *testing.T) {
			t.Run("Initializes default checkpointer state when no checkpointer already exist", func(t *testing.T) {
```

pkg/client/eventsbuilder.go
```diff
@@ -32,14 +32,26 @@ func (builder *eventsBuilder) getStartPosition() *orderer.SeekPosition {
...
-type Checkpointer interface {
+// Checkpoint provides the current position for event processing.
+type Checkpoint interface {
 	// BlockNumber in which the next event is expected.
 	BlockNumber() uint64
 	// TransactionID of the last successfully processed event within the current block.
 	TransactionID() string
 }
 
+// Checkpointer provides the current position for event processing.
```
bestbeforetoday, line 43: Checkpointer has different behaviour to Checkpoint. Perhaps:
```suggestion
// Checkpointer allows update of a checkpoint position after events are successfully processed.
```

pkg/client/eventsbuilder.go
```diff
@@ -32,14 +32,26 @@ func (builder *eventsBuilder) getStartPosition() *orderer.SeekPosition {
...
+type Checkpointer interface {
+	// CheckpointBlock checkpoints the block number.
+	CheckpointBlock(uint64) error
+	// CheckpointTransaction checkpoints the transaction within a block.
+	CheckpointTransaction(uint64, string) error
+	// CheckpointChaincodeEvent checkpoints the event
```
bestbeforetoday, line 49: Need to be well-formed statement, with ending punctuation.
```suggestion
	// CheckpointChaincodeEvent checkpoints the chaincode event.
```

pkg/client/eventsbuilder.go
```diff
@@ -55,11 +67,11 @@ func WithStartBlock(blockNumber uint64) eventOption {
 }
 
 // WithCheckpointer reads events starting at the position recorded by the checkpointer.
```
bestbeforetoday, line 69: Might need a tweak to align with the changed function signature.
```suggestion
// WithCheckpointer reads events starting at the checkpoint position.
```

pkg/client/fileCheckpointer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+type checkpointState struct {
+	BlockNumber   uint64 `json:"blockNumber"`
+	TransactionID string `json:"transactionId"`
+}
+
+func NewFileCheckpointer(path string) (*FileCheckpointer, error) {
+	fileCheckpointer := new(FileCheckpointer)
+	fileCheckpointer.path = path
```
bestbeforetoday, line 28: What's there is perfectly valid. This syntax is more typical. Up to you if you want to change it.

```suggestion
	fileCheckpointer := &FileCheckpointer{
		path: path,
	}
```

pkg/client/fileCheckpointer.go
```diff
@@ -0,0 +1,118 @@
...
+
+	err := fileCheckpointer.loadFromFile()
+	if err != nil {
+		return fileCheckpointer, err
+	}
+
+	err = fileCheckpointer.saveToFile()
+	return fileCheckpointer, err
```
bestbeforetoday, line 36: We typically wouldn't return a result along with an error. The error indicates we failed to create a result.
```suggestion
	if err := fileCheckpointer.loadFromFile(); err != nil {
		return nil, err
	}

	if err := fileCheckpointer.saveToFile(); err != nil {
		return nil, err
	}
	
	return fileCheckpointer, nil
```

pkg/client/fileCheckpointer.go
```diff
@@ -0,0 +1,118 @@
...
+func (c *FileCheckpointer) TransactionID() string {
+	return c.transactionID
+}
+
+func (c *FileCheckpointer) State() *FileCheckpointer {
+	return &FileCheckpointer{
+		blockNumber:   c.blockNumber,
+		transactionID: c.transactionID,
+	}
+}
```
bestbeforetoday, line 67: Something looks wrong here. This method shouldn't be public and I don't think it should be here at all. It allows a FileCheckpointer to return an invalid FileCheckpointer (missing a `path`).

pkg/client/fileCheckpointer.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err := os.Stat(c.path)
+	return !os.IsNotExist(err)
+}
+
+func (c *FileCheckpointer) saveToFile() error {
+	fileCheckpointer := c.State()
+	data, err := json.Marshal(checkpointState{
+		BlockNumber:   fileCheckpointer.blockNumber,
+		TransactionID: fileCheckpointer.transactionID,
+	})
```
bestbeforetoday, line 110:
```suggestion
	data, err := json.Marshal(checkpointState{
		BlockNumber:   c.blockNumber,
		TransactionID: c.transactionID,
	})
```

pkg/client/fileCheckpointer.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+	if err = os.WriteFile(c.path, data, 0600); err != nil {
+		return err
+	}
+	return nil
```
bestbeforetoday, line 117: If the last call in a function already returns what you want, you can just return it directly.
```suggestion
	return os.WriteFile(c.path, data, 0600)
```

pkg/client/fileCheckpointer_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+const checkpointerFilePath = "file_checkpointer_test_file.json"
+
+func after(t *testing.T) {
+	if CheckFileExist(checkpointerFilePath) {
+		err := os.Remove(checkpointerFilePath)
+		require.NoError(t, err)
+	}
+
+}
```
bestbeforetoday, line 24: Similar to `checkpointer_test`, instead of creating and then deleting a file for every test, it might be easier to create a temp directory up-front, use a new temp file within that directory for each new checkpointer, then delete the whole temp directory at the end.

pkg/client/fileCheckpointer_test.go
```diff
@@ -0,0 +1,76 @@
...
+		after(t)
+	})
+
+	t.Run("throws on unwritable file location", func(t *testing.T) {
+
+		badPath := "MISSING_DIRECTORY/test_file.json"
+
+		_, err := NewFileCheckpointer(badPath)
+
+		require.EqualErrorf(t, err, "open MISSING_DIRECTORY/test_file.json: no such file or directory", "Error should be: %v, got: %v", "open MISSING_DIRECTORY/checkpointer.go: no such file or directory", err)
```
bestbeforetoday, line 74: I would only test for what you know will be there, otherwise your tests can be brittle and break if something outside your control (or that isn't really part of the functional spec) changes. The implementation just passes back whatever error the Go standard libraries generate so we should probably either:

1. Just check that an error occurs, not what it looks like (as we don't know); or
2. Throw our own error in the FileCheckpointer implementation that has a sensible message including the file name, and wraps the underlying error. Then just test for an error containing the file name, since we know it will be there (because we put it there) and we don't really care whether somebody comes along later and improves the rest of the error message as long as it still contains the file name.

bestbeforetoday (commented):
I've left several comments but it's all minor code hygiene tweaks. The general content and tests all look good.

bestbeforetoday (approved):

PR #417, opened 2022-03-25
WIP:Checkpointer implemention in java 

sapthasurendran:

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -181,8 +181,14 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.8.5</version>
+            <version>2.9.0</version>
         </dependency>
+        <dependency>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
+            <version>1.1.4</version>
+        </dependency>
+
     </dependencies>
 
     <build>
@@ -378,7 +384,7 @@
                         </goals>
                     </execution>
                 </executions>
-            </plugin>
+            </plugin><plugin><groupId>org.apache.maven.plugins</groupId><artifactId>maven-compiler-plugin</artifactId><configuration><source>8</source><target>8</target></configuration></plugin>
         </plugins>
     </build>
 
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -52,6 +52,16 @@
                 <enabled>false</enabled>
             </snapshots>
         </repository>
+<!--        <repository>-->
+<!--            <id>maven</id>-->
+<!--            <url>https://mvnrepository.com/artifact/</url>-->
+<!--            <releases>-->
+<!--                <enabled>true</enabled>-->
+<!--            </releases>-->
+<!--            <snapshots>-->
+<!--                <enabled>false</enabled>-->
+<!--            </snapshots>-->
+<!--        </repository>-->
     </repositories>
 
     <dependencyManagement>
@@ -167,6 +177,11 @@
         <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.5</version>
         </dependency>
     </dependencies>
 
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -52,16 +52,6 @@
                 <enabled>false</enabled>
             </snapshots>
         </repository>
-<!--        <repository>-->
-<!--            <id>maven</id>-->
-<!--            <url>https://mvnrepository.com/artifact/</url>-->
-<!--            <releases>-->
-<!--                <enabled>true</enabled>-->
-<!--            </releases>-->
-<!--            <snapshots>-->
-<!--                <enabled>false</enabled>-->
-<!--            </snapshots>-->
-<!--        </repository>-->
     </repositories>
 
     <dependencyManagement>
@@ -183,12 +173,6 @@
             <artifactId>gson</artifactId>
             <version>2.9.0</version>
         </dependency>
-        <dependency>
-            <groupId>javax.json</groupId>
-            <artifactId>javax.json-api</artifactId>
-            <version>1.1.4</version>
-        </dependency>
-
     </dependencies>
 
     <build>
@@ -384,7 +368,7 @@
                         </goals>
                     </execution>
                 </executions>
-            </plugin><plugin><groupId>org.apache.maven.plugins</groupId><artifactId>maven-compiler-plugin</artifactId><configuration><source>8</source><target>8</target></configuration></plugin>
+            </plugin>
         </plugins>
     </build>
 
--- a/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
@@ -7,6 +7,8 @@
 package org.hyperledger.fabric.client;
 
 import java.util.Objects;
+import java.util.Optional;
+
 import com.google.protobuf.ByteString;
 import org.hyperledger.fabric.protos.gateway.SignedChaincodeEventsRequest;
 
@@ -38,14 +40,12 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
     @Override
     public ChaincodeEventsRequest.Builder checkpoint(final Checkpoint checkpoint) {
         long blockNumber = checkpoint.getBlockNumber();
-        Boolean transactionStatus = checkpoint.getTransactionId().isPresent();
-        if (blockNumber == 0 && !transactionStatus) {
+        Optional<String> transactionId = checkpoint.getTransactionId();
+        if (blockNumber == 0 && !transactionId.isPresent()) {
             return this;
         }
         startPositionBuilder.startBlock(blockNumber);
-        if (transactionStatus) {
-            this.afterTransactionId = checkpoint.getTransactionId().get();
-        }
+        this.afterTransactionId = transactionId.orElse(null);
         return this;
     }
 
--- a/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
@@ -17,6 +17,7 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
     private final String channelName;
     private final String chaincodeName;
     private final StartPositionBuilder startPositionBuilder = new StartPositionBuilder();
+    private String afterTransactionId;
 
     ChaincodeEventsBuilder(final GatewayClient client, final SigningIdentity signingIdentity, final String channelName,
                            final String chaincodeName) {
@@ -32,6 +33,14 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
     @Override
     public ChaincodeEventsRequest.Builder startBlock(final long blockNumber) {
         startPositionBuilder.startBlock(blockNumber);
+        return this;
+    }
+
+    @Override
+    public ChaincodeEventsRequest.Builder checkpoint(final Checkpoint checkpoint) {
+        startPositionBuilder.startBlock(checkpoint.getBlockNumber());
+        this.afterTransactionId = checkpoint.getTransactionId();
+
         return this;
     }
 
@@ -55,6 +64,7 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
                 .setChaincodeId(chaincodeName)
                 .setIdentity(creator)
                 .setStartPosition(startPositionBuilder.build())
+                .setAfterTransactionId(afterTransactionId)
                 .build();
     }
 }
--- a/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
@@ -7,7 +7,6 @@
 package org.hyperledger.fabric.client;
 
 import java.util.Objects;
-
 import com.google.protobuf.ByteString;
 import org.hyperledger.fabric.protos.gateway.SignedChaincodeEventsRequest;
 
@@ -17,7 +16,7 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
     private final String channelName;
     private final String chaincodeName;
     private final StartPositionBuilder startPositionBuilder = new StartPositionBuilder();
-    private String afterTransactionId = "";
+    private String afterTransactionId;
 
     ChaincodeEventsBuilder(final GatewayClient client, final SigningIdentity signingIdentity, final String channelName,
                            final String chaincodeName) {
@@ -38,13 +37,15 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
 
     @Override
     public ChaincodeEventsRequest.Builder checkpoint(final Checkpoint checkpoint) {
-        Long blockNumber = checkpoint.getBlockNumber();
-        String transactionId = checkpoint.getTransactionId();
-        if (blockNumber == 0 && transactionId.length() == 0) {
+        long blockNumber = checkpoint.getBlockNumber();
+        Boolean transactionStatus = checkpoint.getTransactionId().isPresent();
+        if (blockNumber == 0 && !transactionStatus) {
             return this;
         }
         startPositionBuilder.startBlock(blockNumber);
-        this.afterTransactionId = transactionId;
+        if (transactionStatus) {
+            this.afterTransactionId = checkpoint.getTransactionId().get();
+        }
         return this;
     }
 
@@ -63,12 +64,14 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
 
     private org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest newChaincodeEventsRequestProto() {
         ByteString creator = ByteString.copyFrom(signingIdentity.getCreator());
-        return org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest.newBuilder()
+        org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest.Builder builder = org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest.newBuilder()
                 .setChannelId(channelName)
                 .setChaincodeId(chaincodeName)
                 .setIdentity(creator)
-                .setStartPosition(startPositionBuilder.build())
-                .setAfterTransactionId(afterTransactionId)
-                .build();
+                .setStartPosition(startPositionBuilder.build());
+                if (afterTransactionId != null) {
+                    builder.setAfterTransactionId(afterTransactionId);
+                 }
+        return builder.build();
     }
 }
--- a/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
@@ -17,7 +17,7 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
     private final String channelName;
     private final String chaincodeName;
     private final StartPositionBuilder startPositionBuilder = new StartPositionBuilder();
-    private String afterTransactionId;
+    private String afterTransactionId = "";
 
     ChaincodeEventsBuilder(final GatewayClient client, final SigningIdentity signingIdentity, final String channelName,
                            final String chaincodeName) {
@@ -38,9 +38,13 @@ final class ChaincodeEventsBuilder implements ChaincodeEventsRequest.Builder {
 
     @Override
     public ChaincodeEventsRequest.Builder checkpoint(final Checkpoint checkpoint) {
-        startPositionBuilder.startBlock(checkpoint.getBlockNumber());
-        this.afterTransactionId = checkpoint.getTransactionId();
-
+        Long blockNumber = checkpoint.getBlockNumber();
+        String transactionId = checkpoint.getTransactionId();
+        if (blockNumber == 0 && transactionId.length() == 0) {
+            return this;
+        }
+        startPositionBuilder.startBlock(blockNumber);
+        this.afterTransactionId = transactionId;
         return this;
     }
 
--- a/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
@@ -11,14 +11,14 @@ package org.hyperledger.fabric.client;
  */
 public interface Checkpoint {
     /**
-     *
-     * @return  BlockNumber in which the next event is expected.
+     * The block number in which the next event is expected.
+     * @return  A ledger block number.
      */
     long getBlockNumber();
 
     /**
-     *
-     * @return Transaction ID of the last successfully processed event within the current block.
+     * Transaction Id of the last successfully processed event within the current block.
+     * @return A transaction Id.
      */
     String getTransactionId();
 }
--- a/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
@@ -6,6 +6,8 @@
 
 package org.hyperledger.fabric.client;
 
+import java.util.Optional;
+
 /**
  * Checkpoint provides the current position for event processing.
  */
@@ -20,5 +22,5 @@ public interface Checkpoint {
      * Transaction Id of the last successfully processed event within the current block.
      * @return A transaction Id.
      */
-    String getTransactionId();
+    Optional<String> getTransactionId();
 }
--- a/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2022 IBM All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.hyperledger.fabric.client;
+
+/**
+ * Checkpoint provides the current position for event processing.
+ */
+public interface Checkpoint {
+    /**
+     *
+     * @return  BlockNumber in which the next event is expected.
+     */
+    long getBlockNumber();
+
+    /**
+     *
+     * @return Transaction ID of the last successfully processed event within the current block.
+     */
+    String getTransactionId();
+}
--- a/java/src/main/java/org/hyperledger/fabric/client/Checkpointer.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Checkpointer.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2022 IBM All Rights Reserved.
+ *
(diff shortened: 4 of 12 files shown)


java/pom.xml
```diff
@@ -168,6 +178,11 @@
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.5</version>
```
bestbeforetoday, line 184: Current version is `2.9.0`. Let's not start with a back-level version.

java/src/main/java/org/hyperledger/fabric/client/Checkpoint.java
```diff
@@ -0,0 +1,24 @@
...
+
+package org.hyperledger.fabric.client;
+
+/**
+ * Checkpoint provides the current position for event processing.
+ */
+public interface Checkpoint {
+    /**
+     *
+     * @return  BlockNumber in which the next event is expected.
```
bestbeforetoday, line 15: Needs some actual description.
```suggestion
     * The block number in which the next event is expected.
     * @return A ledger block number.
```

java/src/main/java/org/hyperledger/fabric/client/Checkpointer.java
```diff
@@ -0,0 +1,36 @@
...
+package org.hyperledger.fabric.client;
+
+/**
+ * Checkpointer allows update of a checkpoint position after events are successfully processed.
+ */
+public interface Checkpointer extends Checkpoint {
+
+     /**
+      * CheckpointBlock checkpoints block.
+      * @param blockNumber
```
bestbeforetoday, line 16: Need both method and parameter descriptions.
```suggestion
      * Checkpoint a successfully processed block.
      * <p>Note that the block number is an unsigned 64-bit integer, with the sign bit used to hold the top bit of
      * the number.</p>
      * @param blockNumber a ledger block number.
```

java/src/main/java/org/hyperledger/fabric/client/FileCheckpointer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private void init() throws Exception {
+        this.loadFromFile();
+        this.saveToFile();
+    }
+
+    /**
+     * CheckpointBlock checkpoints the block.
+     * @param blockNumber
+     */
```
bestbeforetoday, line 53: For methods that provide an implementation for something defined in the interface, always use the `@Override` annotation so the compiler can check you are correctly overriding and not accidentally overloading the parent method definition. It also mean you pick up the parent Javadoc for free, so no need to repeat it here.
```suggestion
    @Override
```

java/src/main/java/org/hyperledger/fabric/client/Checkpointer.java
```diff
@@ -0,0 +1,36 @@
...
+      */
+      void  checkpointBlock(long blockNumber) throws Exception;
```
bestbeforetoday, line 18: Avoid throwing the base Exception class. Be more specific about the type of exception that be thrown whenever practical, so calling code can differentiate between different error conditions from different pieces of code. The error here will be in persisting the checkpoint state, so will be [IOException](https://docs.oracle.com/javase/8/docs/api/java/io/IOException.html).
```suggestion
      void  checkpointBlock(long blockNumber) throws IOException;
```

java/src/main/java/org/hyperledger/fabric/client/FileCheckpointer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public long getBlockNumber() {
+        return blockNumber;
+    }
+}
+
+/**
+ * FileCheckpointer to store checkpointer state during file read write operations.
+ */
+public final class FileCheckpointer implements Checkpointer {
```
bestbeforetoday, line 34: You might want to steal the file locking behaviour from https://github.com/hyperledger/fabric-gateway-java/blob/main/src/main/java/org/hyperledger/fabric/gateway/impl/FileCheckpointer.java

bestbeforetoday: It might be helpful to include a `toString()` implementation (see the old SDKs checkpointer implementation)

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public abstract  class CommonCheckpointerTest {
+
+    Checkpointer checkpointerInstance;
+
+    void assertData(Checkpointer checkpointer, Long blockNumber, String transactionId){
```
bestbeforetoday, line 17:
```suggestion
    void assertData(Checkpointer checkpointer, long blockNumber, String transactionId){
```

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,70 @@
...
+    protected abstract void tearDown() throws IOException;
+
+    @Test
+    void initial_checkpointer_state() throws IOException {
+        assertData(this.checkpointerInstance ,Long.valueOf(0),"");
+    }
+
+    @Test
+    void CheckpointBlock_sets_next_block_number_and_empty_transaction_id() throws IOException {
+        Long blockNumber = Long.valueOf(101);
```
bestbeforetoday, line 42:
```suggestion
        long blockNumber = 101;
```

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        this.checkpointerInstance.checkpointBlock(blockNumber);
+        assertData(this.checkpointerInstance ,blockNumber +1,"");
+    }
+
+    @Test
+    void checkpointTransaction_sets_block_number_and_transaction_id() throws Exception {
+        Long blockNumber = Long.valueOf(101);
```
bestbeforetoday, line 50:
```suggestion
        long blockNumber = 101;
```

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,70 @@
...
+    Checkpointer checkpointerInstance;
+
+    void assertData(Checkpointer checkpointer, Long blockNumber, String transactionId){
+        assertThat(checkpointer.getBlockNumber()).isEqualTo(blockNumber);
+        assertThat(checkpointer.getTransactionId()).isEqualTo(transactionId);
+    }
+
+    @BeforeEach
+    void beforeEach() throws IOException {
+        this.checkpointerInstance = getCheckpointerInstance();
```
bestbeforetoday, line 24:
```suggestion
       checkpointerInstance = getCheckpointerInstance();
```

java/src/main/java/org/hyperledger/fabric/client/FileCheckpointer.java
```diff
@@ -0,0 +1,169 @@
...
+    private static final Set<OpenOption> OPEN_OPTIONS = Collections.unmodifiableSet(EnumSet.of(
+            StandardOpenOption.CREATE,
+            StandardOpenOption.READ,
+            StandardOpenOption.WRITE
+    ));
+    private final FileChannel fileChannel;
+    private static final String CONFIG_KEY_BLOCK = "blockNumber";
+    private static final String CONFIG_KEY_TRANSACTIONID = "transactionId";
+
+    FileCheckpointer(final String path) throws IOException {
```
bestbeforetoday, line 51: The constructor should take a [Path](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html) instead of a String parameter.

java/src/test/java/org/hyperledger/fabric/client/InMemoryCheckpointerTest.java
```diff
@@ -0,0 +1,19 @@
+package org.hyperledger.fabric.client;
+
+
+import java.io.IOException;
+
+public class InMemoryCheckpointerTest extends CommonCheckpointerTest {
+    InMemoryCheckpointer inMemoryCheckpointer;
```
bestbeforetoday, line 7: This is never read so can be removed

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -138,6 +138,161 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+        assertThat(request.getChaincodeId()).isEqualTo("CHAINCODE_NAME");
+
+        assertThat(request.getStartPosition().getTypeCase()).isEqualTo(Ab.SeekPosition.TypeCase.SPECIFIED);
+        assertThat(request.getStartPosition().getSpecified().getNumber()).isEqualTo(blockNumber);
+        assertThat(request.getAfterTransactionId()).isEqualTo(transactionId);
+
+    }
+
+    @Test
+    void sends_valid_request_with_with_start_block_and_checkpointer_chaincode_event() throws  Exception {
```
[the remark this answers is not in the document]
sapthasurendran: To add to the code coverage this would be useful right?

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,70 @@
...
+    void afterEach() throws IOException {
+        tearDown();
+    }
+
+    protected abstract Checkpointer getCheckpointerInstance() throws IOException;
+    protected abstract void tearDown() throws IOException;
+
+    @Test
+    void initial_checkpointer_state() throws IOException {
+        assertData(this.checkpointerInstance ,Long.valueOf(0),"");
```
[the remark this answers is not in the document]
sapthasurendran: Long.valueOf still need to be there right ?

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -57,6 +74,18 @@ void afterEach() {
...
+        assertThat(actual.getChannelId()).isEqualTo(expectedRequest.channelName);
+        assertThat(actual.getChaincodeId()).isEqualTo(expectedRequest.chaincodeName);
+        assertThat(actual.getStartPosition().getTypeCase()).isEqualTo(expectedRequest.typeCase);
+        if(expectedRequest.blockNumber.isPresent()) {
+            assertThat(actual.getStartPosition().getSpecified().getNumber()).isEqualTo(expectedRequest.blockNumber.get());
+        }
+        if(expectedRequest.transactionId.isPresent()) {
+            assertThat(actual.getAfterTransactionId()).isEqualTo(expectedRequest.transactionId.get());
+        }
+    }
```
bestbeforetoday, line 87: I would be tempted to remove the assertions of channel and chaincode name for the tests that are focusing on start position and checkpoint. If they are already covered adequately in earlier tests of different permutations where they could be missed, great. If not, we could just have a specific test for channel and chaincode name so we don't need to worry about them in other tests.

It might be simpler to break what's left up a bit to make it (much) simpler. I _think_ these methods save some copy & paste of code, and the method names make it clearer what is being asserted. See what you think.
```suggestion
    void assertStartPosition(final org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest actual, final long blockNumber, final String transactionId) {
        assertThat(actual.getStartPosition().getTypeCase()).isEqualTo(Ab.SeekPosition.TypeCase.SPECIFIED);
        assertThat(actual.getStartPosition().getSpecified().getNumber()).isEqualTo(blockNumber);
        assertThat(actual.getAfterTransactionId()).isEqualTo(transactionId);
    }

    void assertNextCommit(final org.hyperledger.fabric.protos.gateway.ChaincodeEventsRequest actual) {
        assertThat(actual.getStartPosition().getTypeCase()).isEqualTo(Ab.SeekPosition.TypeCase.NEXT_COMMIT);
    }
```

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+        long startBlock = -1;
+        Checkpointer checkpointer = new InMemoryCheckpointer();
+        EventsRequest<ChaincodeEvent> eventsRequest = network.newChaincodeEventsRequest("CHAINCODE_NAME")
+                .startBlock(startBlock)
+                .checkpoint(checkpointer)
+                .build();
+
+        try (CloseableIterator<ChaincodeEvent> iter = eventsRequest.getEvents()) {
+            iter.hasNext(); // Interact with iterator before asserting to ensure async request has been made
+        }
```
bestbeforetoday, line 172: This gets duplicated a lot. It might be worth having a utility method to do this in a one-line call. Up to you. Not a big deal either way.

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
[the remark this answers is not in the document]
bestbeforetoday: Sure, that's fine

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+
+        SignedChaincodeEventsRequest signedRequest = mocker.captureChaincodeEvents();
+        ChaincodeEventsRequest request = ChaincodeEventsRequest.parseFrom(signedRequest.getRequest());
+
+        assertChaincodeRequest(request, new ExpectedRequest("NETWORK", "CHAINCODE_NAME", Ab.SeekPosition.TypeCase.SPECIFIED, Optional.ofNullable(checkpointer.getBlockNumber()),checkpointer.getTransactionId()));
+    }
+
+    @Test
+    void sends_valid_request_with_with_start_block_and_checkpointer_chaincode_event() throws  Exception {
```
bestbeforetoday, line 261:
```suggestion
    void uses_checkpointed_chaincode_event_block_and_transaction() throws  Exception {
```

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+        ChaincodeEventsRequest request = ChaincodeEventsRequest.parseFrom(signedRequest.getRequest());
+
+        assertChaincodeRequest(request, new ExpectedRequest("NETWORK","CHAINCODE_NAME",Ab.SeekPosition.TypeCase.SPECIFIED, Optional.ofNullable(checkpointer.getBlockNumber()),Optional.empty()));
+    }
+
+    @Test
+    void uses_checkpoint_block_and_transaction_instead_of_specified_start_block() throws Exception {
+        long startBlock = -1;
+        long blockNumber = 101;
+        String transactionId = "TRANSACTION_ID";
```
bestbeforetoday, line 205: These variables only get used once so probably not needed

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+        SignedChaincodeEventsRequest signedRequest = mocker.captureChaincodeEvents();
+        ChaincodeEventsRequest request = ChaincodeEventsRequest.parseFrom(signedRequest.getRequest());
+
+        assertChaincodeRequest(request, new ExpectedRequest("NETWORK", "CHAINCODE_NAME", Ab.SeekPosition.TypeCase.NEXT_COMMIT, Optional.empty(),Optional.empty()));
+    }
+
+    @Test
+    void uses_checkpoint_block_and_transaction_with_unset_start_block() throws Exception {
+        long blockNumber = 1;
+        String transactionId = "TRANSACTION_ID";
```
bestbeforetoday, line 243: These variables only get used once so probably not needed

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+
+        SignedChaincodeEventsRequest signedRequest = mocker.captureChaincodeEvents();
+        ChaincodeEventsRequest request = ChaincodeEventsRequest.parseFrom(signedRequest.getRequest());
+
+        assertChaincodeRequest(request, new ExpectedRequest("NETWORK", "CHAINCODE_NAME", Ab.SeekPosition.TypeCase.SPECIFIED, Optional.ofNullable(checkpointer.getBlockNumber()),checkpointer.getTransactionId()));
+    }
+
+    @Test
+    void sends_valid_request_with_with_start_block_and_checkpointer_chaincode_event() throws  Exception {
+        long blockNumber = 1;
```
bestbeforetoday, line 262: Once the assertions are refactored, this variable probably isn't necessary

java/src/test/java/org/hyperledger/fabric/client/ChaincodeEventsTest.java
```diff
@@ -131,11 +155,140 @@ void sends_valid_request_with_specified_start_block_number_using_sign_bit_for_un
...
+        ChaincodeEventsRequest request = ChaincodeEventsRequest.parseFrom(signedRequest.getRequest());
+
         assertThat(request.getChannelId()).isEqualTo("NETWORK");
         assertThat(request.getChaincodeId()).isEqualTo("CHAINCODE_NAME");
 
         assertThat(request.getStartPosition().getTypeCase()).isEqualTo(Ab.SeekPosition.TypeCase.SPECIFIED);
-        assertThat(request.getStartPosition().getSpecified().getNumber()).isEqualTo(startBlock);
+        assertThat(request.getStartPosition().getSpecified().getNumber()).isEqualTo(blockNumber);
+        assertThat(request.getAfterTransactionId()).isEqualTo(event.getTxId());
+        assertChaincodeRequest(request, new ExpectedRequest("NETWORK", "CHAINCODE_NAME", Ab.SeekPosition.TypeCase.SPECIFIED, Optional.ofNullable(blockNumber), Optional.of(event.getTxId())));
```
bestbeforetoday, line 290: Lots of duplication here. Needs tidying up and would suggest referencing `chaincodeEvent.getBlockNumber()` and `chaincodeEvent.getTransactionId()` as expected values

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    void assertData(Checkpointer checkpointer, long blockNumber, Optional<String> transactionId) {
+        assertThat(checkpointer.getBlockNumber()).isEqualTo(blockNumber);
+        if (transactionId.isPresent()) {
+            assertTrue(checkpointer.getTransactionId().isPresent());
+            assertThat(checkpointer.getTransactionId().get()).isEqualTo(transactionId.get());
+        } else {
+            assertFalse(checkpointer.getTransactionId().isPresent());
+        }
+    }
```
bestbeforetoday, line 28: I wouldn't pass an Optional in here. AssertJ has better ways of checking an Optional value too. If you want to have assert methods, I'd probably break it up like below with method overloading or two differently named methods, or you might find simpler assertions make it not worth the effort. If sub-classess will use these methods, make them `protected`; otherwise `private`.
```suggestion
    protected assertCheckpoint(final Checkpoint checkpoint, final long blockNumber) {
        assertThat(checkpoint.getBlockNumber()).isEqualTo(blockNumber);
        assertThat(checkpoint.getTransactionId()).isNotPresent();
    }

    protected assertCheckpoint(final Checkpoint checkpoint, final long blockNumber, final String transactionId) {
        assertThat(checkpoint.getBlockNumber()).isEqualTo(blockNumber);
        assertThat(checkpoint.getTransactionId()).isEqualTo(transactionId);
    }
```

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public abstract  class CommonCheckpointerTest {
+
+    Checkpointer checkpointerInstance;
```
bestbeforetoday, line 18: If this is used by subclasses, make it `protected`; otherwise `private`.

java/src/test/java/org/hyperledger/fabric/client/FileCheckpointerTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class FileCheckpointerTest extends CommonCheckpointerTest {
+    FileCheckpointer fileCheckpointer;
+    private static final TestUtils testUtils = TestUtils.getInstance();
+
+    @Override
+    protected FileCheckpointer getCheckpointerInstance() throws IOException {
+        Path file = testUtils.createTempFile(".json");
+        this.fileCheckpointer = new FileCheckpointer(file);
```
bestbeforetoday, line 21: No `this` unless it's necessary because of a shadowed variable name!
```suggestion
       fileCheckpointer = new FileCheckpointer(file);
```

java/src/test/java/org/hyperledger/fabric/client/FileCheckpointerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class FileCheckpointerTest extends CommonCheckpointerTest {
+    FileCheckpointer fileCheckpointer;
```
bestbeforetoday, line 15: Vsibility and we're already in a FileCheckpointerTest so the "file" part of the variable naming is kinda implied.
```suggestion
    private FileCheckpointer checkpointer;
```

java/src/test/java/org/hyperledger/fabric/client/FileCheckpointerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    protected void tearDown() throws IOException {
+        this.fileCheckpointer.close();
+    }
+
+    @Test
+    void state_is_persisted() throws IOException {
+        Path file = testUtils.createTempFile(".json");
+        this.fileCheckpointer = new FileCheckpointer(file);
+        this.fileCheckpointer.checkpointTransaction(1, Optional.ofNullable("TRANSACTION_ID"));
+        this.fileCheckpointer.close();
```
bestbeforetoday, line 35: Let's give try-with-resources a try since that's why we implemented AutoCloseable!
```suggestion
        try (fileCheckpointer = new FileCheckpointer(file)) {
            fileCheckpointer.checkpointTransaction(1, Optional.ofNullable("TRANSACTION_ID"));
        }
```

java/src/test/java/org/hyperledger/fabric/client/InMemoryCheckpointerTest.java
```diff
@@ -0,0 +1,17 @@
+package org.hyperledger.fabric.client;
+
+
+import java.io.IOException;
+
+public class InMemoryCheckpointerTest extends CommonCheckpointerTest {
+
+    @Override
+    protected InMemoryCheckpointer getCheckpointerInstance() throws IOException {
```
bestbeforetoday, line 9: Not a big deal; just to be tidy
```suggestion
    protected InMemoryCheckpointer getCheckpointerInstance() {
```

java/src/test/java/org/hyperledger/fabric/client/InMemoryCheckpointerTest.java
```diff
@@ -0,0 +1,17 @@
...
+        return new InMemoryCheckpointer();
+    }
+
+    @Override
+    protected void tearDown() throws IOException {
```
bestbeforetoday, line 14: Not a big deal; just to be tidy
```suggestion
    protected void tearDown() {
```

bestbeforetoday (changes requested):
Sorry so many comments again! It really is looking good, just odd Java tody-up as part of the learning process

java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
```diff
@@ -35,6 +35,20 @@
...
+    public ChaincodeEventsRequest.Builder checkpoint(final Checkpoint checkpoint) {
+        long blockNumber = checkpoint.getBlockNumber();
+        Boolean transactionStatus = checkpoint.getTransactionId().isPresent();
+        if (blockNumber == 0 && !transactionStatus) {
+            return this;
+        }
+        startPositionBuilder.startBlock(blockNumber);
+        if (transactionStatus) {
+            this.afterTransactionId = checkpoint.getTransactionId().get();
+        }
```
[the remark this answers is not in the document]
sapthasurendran: I think afterTransactionId need to be a function here in order to execute  `transactionId.ifPresent(this::afterTransactionId);`

java/src/main/java/org/hyperledger/fabric/client/FileCheckpointer.java
```diff
@@ -0,0 +1,186 @@
...
+     * Releases the resources and closes the file channel.
+     * @throws IOException
+     */
+
+    public void close() throws IOException {
+        fileChannel.close(); // Also releases lock
+    }
+
+    /**
+     * Confirms the changes made to the file have been written to the storage device.
```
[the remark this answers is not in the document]
sapthasurendran: how exactly the sync has to be used by the client?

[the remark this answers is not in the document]
bestbeforetoday: They would call it immediately after calling one of the checkpoint methods. This is a tradeoff of performance vs. data integrity that they can choose to make. If you don't call sync, your data goes into an I/O buffer somewhere (in the process, the operating system, or storage device). If the system dies this buffer may not have been written to persistent storage and so data could be lost, but it is dependent on the system and any hardware protection to help prevent this. Calling sync makes a best effort to ensure the data is written to persistent storage and not buffered anywhere, but there is a significant performance penalty for doing this. So it's a choice for the user.

java/src/main/java/org/hyperledger/fabric/client/ChaincodeEventsBuilder.java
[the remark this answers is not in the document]
bestbeforetoday: Whoops, you are absolutely correct! I was thinking of the setter on the protobuf builder. Your call on what you prefer I guess. Just store the Optional in `afterTransactionId` and then use `afterTransactionId.ifPresent(builder::setAfterTransactionId)` later, or grab the value out of the Optional here, perhaps using `afterTransactionId = transactionId.orElse(null)`

java/src/test/java/org/hyperledger/fabric/client/CommonCheckpointerTest.java
sapthasurendran: In case of checkpointBlock , should we set  transactionID to "" or null ?
If empty string, then `checkpoint.getTransactionId().isNotPresent() `will fail

bestbeforetoday (approved):

PR #426, opened 2022-04-06
Scenario Test  Chaincode Checkpointer

sapthasurendran:


java/src/test/java/scenario/GatewayContext.java
```diff
@@ -11,11 +11,7 @@
 import java.util.concurrent.TimeUnit;
 
 import io.grpc.ManagedChannel;
-import org.hyperledger.fabric.client.ChaincodeEvent;
-import org.hyperledger.fabric.client.CloseableIterator;
-import org.hyperledger.fabric.client.Contract;
-import org.hyperledger.fabric.client.Gateway;
-import org.hyperledger.fabric.client.Network;
+import org.hyperledger.fabric.client.*;
```
bestbeforetoday, line 14: Generally I would prefer to avoid wildcard imports

scenario/features/chaincodeevents.feature
```diff
@@ -53,3 +53,18 @@ Feature: Chaincode event listening
         And I set the transaction arguments to ["close", "after"]
         And I invoke the transaction
         Then I should receive a chaincode event named "close" with payload "after" on "listener2"
+
+    Scenario: Checkpoint of chaincode events
+        Given I create a checkpointer
+        And I use my checkpointer to listen for chaincode events from basic
```
bestbeforetoday, line 59: To better fit the language used in other steps (e.g. "I invoke **the** transaction"):
```suggestion
        And I use the checkpointer to listen for chaincode events from basic
```

scenario/features/chaincodeevents.feature
```diff
@@ -53,3 +53,18 @@ Feature: Chaincode event listening
...
+        And I prepare to submit an event transaction
+        And I set the transaction arguments to ["checkpoint", "one"]
+        And I invoke the transaction
+        Then I should receive a chaincode event named "checkpoint" with payload "one"
+        Then I should checkpoint the chaincode event
```
bestbeforetoday, line 64: Then steps are assertions and this is not an assertion. It should not be needed either. We already said in an earlier step that we were using a checkpointer to listen for events so checkpointing should happen (in the test implementation code) without an explicit checkpoint step

java/src/test/java/scenario/ScenarioSteps.java
```diff
@@ -74,6 +76,8 @@ public class ScenarioSteps {
     private static final Path DOCKER_COMPOSE_DIR = Paths.get(FIXTURES_DIR.toString(), "docker-compose")
             .toAbsolutePath();
     private static final String DEFAULT_LISTENER_NAME = "";
+    private static Checkpointer checkpointer;
```
bestbeforetoday, line 79: This should not be static as it should not persist across scenarios. The checkpointer should be associated with a specific EventListener instance. The EventListener should call the checkpointer when it is asked for the next event. Perhaps a specific CheckpointEventListener implementation class that looks like the current EventListener but also does the checkpointing.

java/src/test/java/scenario/ScenarioSteps.java
```diff
@@ -74,6 +76,8 @@ public class ScenarioSteps {
...
+    private static ChaincodeEvent lastChaincodeEventReceived;
```
bestbeforetoday, line 80: This should not be needed

java/src/test/java/scenario/ScenarioSteps.java
```diff
@@ -732,4 +753,11 @@ private static PrivateKey getPrivateKey(final Path privateKeyPath) throws IOExce
             return Identities.readPrivateKey(privateKeyReader);
         }
     }
+
+    private void assertCheckpointerExist() {
```
bestbeforetoday, line 757: I don't think the rest of the Java test implementation does explicit checking that things are non-null so I wouldn't bother doing that for this case either

scenario/go/scenario_test.go
```diff
@@ -38,10 +38,12 @@ const (
...
-	gateways                 map[string]*GatewayConnection
-	currentGateway           *GatewayConnection
-	transaction              *Transaction
-	lastCommittedBlockNumber uint64
+	gateways                   map[string]*GatewayConnection
+	currentGateway             *GatewayConnection
+	transaction                *Transaction
+	lastCommittedBlockNumber   uint64
+	checkpointer               *client.InMemoryCheckpointer
+	lastChaincodeEventReceived *client.ChaincodeEvent
```
bestbeforetoday, line 46: This shouldn't be needed as checkpointing should happen as events are read rather than as a discreet step after reading

bestbeforetoday (changes requested):
Generally looks good but I think it needs some structural tweaks. See the comments for details. The Java comments will apply to the TypeScript code too

bestbeforetoday (approved):

PR #427, opened 2022-04-14
Block checkpointer 

sapthasurendran:


sapthasurendran:
For java and go , with the current implementation I don't think it is possible to checkpoint block 0 .

bestbeforetoday:
> For java and go , with the current implementation I don't think it is possible to checkpoint block 0 .

Checkpointing block 0 should set the checkpoint block number to 1 (blockNumber + 1). What issue are you seeing?

sapthasurendran:
> > For java and go , with the current implementation I don't think it is possible to checkpoint block 0 .
> 
> Checkpointing block 0 should set the checkpoint block number to 1 (blockNumber + 1). What issue are you seeing?

0 can be a zero value (java and go) , and in case of block checkpointer how can we differentiate zero value since we don't have transaction ID to check for null value ?

bestbeforetoday:
An unset checkpointer would have block number zero and no transaction ID. A checkpointer after checkpointing block zero would have block number one and no transaction ID since checkpointing a block sets the checkpointer to the next block number. You can’t checkpoint a block and still have block number zero in the checkpointer.

scenario/features/blockevents.feature
```diff
@@ -77,3 +77,24 @@ Feature: Block event listening
...
             | block and private data |
+
+    Scenario Outline: Checkpoint of block events
+        Given I create a checkpointer
+        And I use my checkpointer to listen for <type> events
+        And I prepare to submit an echo transaction
+        And I set the transaction arguments to ["checkpoint"]
+        And I invoke the transaction
+        Then I should receive a <type> event
+        Then I should checkpoint the <type> event
```
bestbeforetoday, line 88: Similar to the chaincode checkpointing PR, we shouldn't have a separate step to do the checkpointing. Since we used a checkpointer to listen for events we should be checkpointing events read from that listener.

node/src/blockevents.test.ts
```diff
@@ -237,15 +255,66 @@ describe('Block Events', () => {
...
-            expect(stop).toBeDefined();
-            expect(stop?.getTypeCase()).toBe(orderer.SeekPosition.TypeCase.SPECIFIED);
-            expect(stop?.getSpecified()).toBeDefined();
-            expect(stop?.getSpecified()?.getNumber()).toBe(Number.MAX_SAFE_INTEGER);
+
+            assertStartPositionToBeSpecified(seekInfo, Number(startBlock));
+            assertStopPosition(seekInfo);
+        });
+
+        it('Sends valid request with specified start block number and fresh checkpointer', async () => {
```
bestbeforetoday, line 263: The Java unit tests cover 4 cases (whereas here are only 3), and the Java test naming better describes the expected behaviour. It's probably worth changing tweaking these tests to match the Java ones:

1. Uses specified start block instead of unset checkpoint
2. Uses checkpoint block instead of specified start block
3. Uses default start position with unset checkpoint and no start block
4. Uses checkpoint block zero with set transaction ID instead of specified start block

pkg/client/blockevents_test.go
```diff
@@ -143,6 +143,146 @@ func TestBlockEvents(t *testing.T) {
 		test.AssertProtoEqual(t, expected, actual)
 	})
 
+	t.Run("Sends valid request with specified start block number and fresh checkpointer", func(t *testing.T) {
```
bestbeforetoday, line 146: The Java unit tests cover 4 cases (whereas here are only 3), and the Java test naming better describes the expected behaviour. It's probably worth changing tweaking these tests to match the Java ones:

1. Uses specified start block instead of unset checkpoint
2. Uses checkpoint block instead of specified start block
3. Uses default start position with unset checkpoint and no start block
4. Uses checkpoint block zero with set transaction ID instead of specified start block

scenario/features/blockevents.feature
```diff
@@ -77,3 +77,23 @@ Feature: Block event listening
             | block                  |
             | filtered block         |
             | block and private data |
+
+    Scenario Outline: Checkpoint of block events
+        Given I create a checkpointer
+        And I use my checkpointer to listen for <type> events
```
bestbeforetoday, line 83: All the other steps use "a"/"an" or "the", not "my". It's a small thing but I think it's worth being consistent.
```suggestion
        And I use the checkpointer to listen for <type> events
```
Remember the code step definitions will need a matching change in all three languages too.

scenario/features/blockevents.feature
```diff
@@ -77,3 +77,23 @@ Feature: Block event listening
...
+        And I prepare to submit an echo transaction
+        And I set the transaction arguments to ["checkpoint"]
+        And I invoke the transaction
+        Then I should receive a <type> event
+        When I stop listening for <type> events
+        And I prepare to submit an echo transaction
+        And I set the transaction arguments to ["echo"]
+        And I invoke the transaction
+        And I use my checkpointer to listen for <type> events
```
bestbeforetoday, line 92:
```suggestion
        And I use the checkpointer to listen for <type> events
```

scenario/go/scenario_test.go
```diff
@@ -38,10 +40,14 @@ const (
...
-	transaction              *Transaction
-	lastCommittedBlockNumber uint64
+	gateways                        map[string]*GatewayConnection
+	currentGateway                  *GatewayConnection
+	transaction                     *Transaction
+	lastCommittedBlockNumber        uint64
+	checkpointer                    *client.InMemoryCheckpointer
+	lastBlockEventReceived          *common.Block
+	lastFilteredBlockEventReceived  *peer.FilteredBlock
+	lastBlockAndPrivateDataReceived *peer.BlockAndPrivateData
```
bestbeforetoday, line 50: `checkpointer` is stored on the GatewayConnection and the last block received are not used at all now that checkpointing is taken care of in the listener. These should all be removed.

scenario/node/src/customworld.ts
```diff
@@ -218,10 +218,23 @@ export class CustomWorld {
...
+
+    async listenForBlockAndPrivateDataEventsUsingCheckpointer(listenerName: string): Promise<void> {
+        await this.getCurrentGateway().listenForBlockAndPrivateDataEventsUsingCheckpointer(listenerName, { checkpoint: this.getCurrentGateway().getCheckpointer() });
+    }
+
     async replayBlockEvents(listenerName: string, startBlock: bigint): Promise<void> {
         await this.getCurrentGateway().listenForBlockEvents(listenerName, { startBlock });
     }
 
+
```
bestbeforetoday, line 237: No need for an extra blank line -- remove

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
         const event: ChaincodeEvent = await this.getChaincodeEventListener(listenerName).next();
         return event;
     }
-
```
bestbeforetoday, line 87: You lost a blank line between method definitions. It would be good to have it back for readability

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
     }
 
-    async nextBlockEvent(listenerName: string): Promise<unknown> {
-        return await this.getBlockEventListener(listenerName).next();
+    async listenForBlockEventsUsingCheckpointer(listenerName: string, options?: BlockEventsOptions): Promise<void> {
+        this.closeBlockEvents(listenerName);
+        const events = await this.getNetwork().getBlockEvents(options);
+        const listener = new CheckpointEventListener<common.Block>(events, async (event: common.Block): Promise<void> => {
+            const blockNumber = event.getHeader()?.getNumber();
+            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber!));
```
bestbeforetoday, line 99:
```suggestion
            const header = assertDefined(event.getHeader(), 'block header');
            const blockNumber = header.getNumber();
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
+        });
+        this.#blockEventListeners.set(listenerName, listener);
     }
 
+    async nextBlockEvent(listenerName: string): Promise<unknown> {
+        const event  = await this.getBlockEventListener(listenerName).next();
+        return event;
```
bestbeforetoday, line 106:
```suggestion
        return await this.getBlockEventListener(listenerName).next();
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
     }
 
-    async nextFilteredBlockEvent(listenerName: string): Promise<unknown> {
-        return await this.getFilteredBlockEventListener(listenerName).next();
+    async listenForFilteredBlockEventsUsingCheckpointer(listenerName: string, options?: BlockEventsOptions): Promise<void> {
+        this.closeFilteredBlockEvents(listenerName);
+        const events = await this.getNetwork().getFilteredBlockEvents(options);
+        const listener = new CheckpointEventListener<peer.FilteredBlock>(events, async (event: peer.FilteredBlock): Promise<void> => {
+            const blockNumber = event.getNumber();
+            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber));
```
bestbeforetoday, line 120:
```suggestion
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
+        });
+        this.#filteredBlockEventListeners.set(listenerName, listener);
     }
 
+    async nextFilteredBlockEvent(listenerName: string): Promise<unknown> {
+        const event = await this.getFilteredBlockEventListener(listenerName).next();
+        return event;
```
bestbeforetoday, line 127:
```suggestion
        return await this.getFilteredBlockEventListener(listenerName).next();
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
         const listener = new BaseEventListener<peer.BlockAndPrivateData>(events);
         this.#blockAndPrivateDataEventListeners.set(listenerName, listener);
     }
 
+    async listenForBlockAndPrivateDataEventsUsingCheckpointer(listenerName: string, options?: BlockEventsOptions): Promise<void> {
+        this.closeBlockAndPrivateDataEvents(listenerName);
+        const events = await this.getNetwork().getBlockAndPrivateDataEvents(options);
+        const listener = new CheckpointEventListener<peer.BlockAndPrivateData>(events, async (event: peer.BlockAndPrivateData): Promise<void>  => {
+            const blockNumber = event.getBlock()?.getHeader()?.getNumber();
+            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber!));
```
bestbeforetoday, line 141:
```suggestion
            const block = assertDefined(event.getBlock(), 'block');
            const header = assertDefined(block.getHeader(), 'block header');
            const blockNumber = header.getNumber();
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -84,38 +84,68 @@ export class GatewayContext {
...
+        });
+        this.#blockAndPrivateDataEventListeners.set(listenerName, listener);
+    }
+
     async nextBlockAndPrivateDataEvent(listenerName: string): Promise<unknown> {
-        return await this.getBlockAndPrivateDataEventListener(listenerName).next();
+        const event = await this.getBlockAndPrivateDataEventListener(listenerName).next();
+        return event;
```
bestbeforetoday, line 148:
```suggestion
        return await this.getBlockAndPrivateDataEventListener(listenerName).next();
```

bestbeforetoday (changes requested):

scenario/go/scenario_test.go
```diff
@@ -40,14 +40,13 @@ const (
...
-	lastCommittedBlockNumber        uint64
-	checkpointer                    *client.InMemoryCheckpointer
-	lastBlockEventReceived          *common.Block
-	lastFilteredBlockEventReceived  *peer.FilteredBlock
-	lastBlockAndPrivateDataReceived *peer.BlockAndPrivateData
+	gateways                       map[string]*GatewayConnection
+	currentGateway                 *GatewayConnection
+	transaction                    *Transaction
+	lastCommittedBlockNumber       uint64
+	checkpointer                   *client.InMemoryCheckpointer
```
bestbeforetoday, line 47: Not used -- remove

scenario/go/scenario_test.go
```diff
@@ -40,14 +40,13 @@ const (
...
+	lastBlockEventReceived         *common.Block
```
bestbeforetoday, line 48: Not used -- remove

scenario/go/scenario_test.go
```diff
@@ -40,14 +40,13 @@ const (
...
+	lastFilteredBlockEventReceived *peer.FilteredBlock
```
bestbeforetoday, line 49: Not used -- remove

scenario/node/src/gatewaycontext.ts
```diff
@@ -95,16 +96,17 @@ export class GatewayContext {
         this.closeBlockEvents(listenerName);
         const events = await this.getNetwork().getBlockEvents(options);
         const listener = new CheckpointEventListener<common.Block>(events, async (event: common.Block): Promise<void> => {
-            const blockNumber = event.getHeader()?.getNumber();
-            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber!));
+            const header = assertDefined(event.getHeader(), 'block header');
+            const blockNumber = header.getNumber();
+            await this.getCheckpointer()?.checkpointBlock(BigInt(blockNumber));
```
bestbeforetoday, line 101: The point of using `getCheckpointer()` is that it guarantees a checkpointer is returned since it throws if one is not present. The reason this isn't obvious is because the implementation and return type of `getCheckpointer()` is wrong. It should be

```typescript
    private getCheckpointer(): Checkpointer {
        return assertDefined(this.#checkpointer, 'checkpointer');
    }
```

There is no need to use the `?` operator when the value can't be undefined.
```suggestion
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -117,7 +119,7 @@ export class GatewayContext {
         const events = await this.getNetwork().getFilteredBlockEvents(options);
         const listener = new CheckpointEventListener<peer.FilteredBlock>(events, async (event: peer.FilteredBlock): Promise<void> => {
             const blockNumber = event.getNumber();
-            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber));
+            await this.getCheckpointer()?.checkpointBlock(BigInt(blockNumber));
```
bestbeforetoday, line 122:
```suggestion
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

scenario/node/src/gatewaycontext.ts
```diff
@@ -137,15 +139,16 @@ export class GatewayContext {
         this.closeBlockAndPrivateDataEvents(listenerName);
         const events = await this.getNetwork().getBlockAndPrivateDataEvents(options);
         const listener = new CheckpointEventListener<peer.BlockAndPrivateData>(events, async (event: peer.BlockAndPrivateData): Promise<void>  => {
-            const blockNumber = event.getBlock()?.getHeader()?.getNumber();
-            await this.#checkpointer?.checkpointBlock(BigInt(blockNumber!));
+            const block = assertDefined(event.getBlock(), 'block');
+            const header = assertDefined(block.getHeader(), 'block header');
+            const blockNumber = header.getNumber();
+            await this.getCheckpointer()?.checkpointBlock(BigInt(blockNumber));
```
bestbeforetoday, line 145:
```suggestion
            await this.getCheckpointer().checkpointBlock(BigInt(blockNumber));
```

bestbeforetoday (changes requested):
Just a few minor tweaks required and it's done.

Issue #394, opened 2022-02-17
Persistence of proposal, transaction and commit objects


bestbeforetoday:
**As an** application developer
**I want** to persist proposal, transaction and commit objects
**So that** I can resume a submit flow following an application restart

This is already possible using the off-line signing flow but the methods for re-creating these objects from serialized bytes do not make it obvious that this is the case, either through their naming or the presence of a signature parameter, which must be left empty for the persistence flow without off-line signing.

Suggest simply adding methods such as `newUnsignedProposal(bytes)` to Gateway to support this flow. The implementation would be largely common with the `newSignedProposal(bytes, signature)` methods.

PR #435, opened 2022-05-24
Persist Proposal, Transaction, Commit and Event Objects (Node JS)

sapthasurendran:
Contributes to  #394 

### DIFF
--- a/node/src/chaincodeeventsrequest.ts
+++ b/node/src/chaincodeeventsrequest.ts
@@ -39,7 +39,7 @@ export interface ChaincodeEventsRequest extends Signable {
 export interface ChaincodeEventsRequestOptions {
     client: GatewayClient;
     signingIdentity: SigningIdentity;
-    request: gateway.ChaincodeEventsRequest;
+    signedRequest: gateway.SignedChaincodeEventsRequest;
 }
 
 export class ChaincodeEventsRequestImpl implements ChaincodeEventsRequest {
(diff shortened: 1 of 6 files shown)


denyeart:
@sapthasurendran If you add text like "Resolves #394" or "Closes #394" in the PR and commit message it will automatically get associated with the GitHub issue.

node/src/chaincodeeventsrequest.ts
```diff
@@ -36,22 +36,21 @@ export interface ChaincodeEventsRequest extends Signable {
     getEvents(options?: CallOptions): Promise<CloseableAsyncIterable<ChaincodeEvent>>;
 }
 
-export interface ChaincodeEventsRequestOptions {
+export interface SignedChaincodeEventsRequestOptions {
```
bestbeforetoday, line 39: These are still the options used when creating a ChaincodeEventsRequest so the name should remain the same.

node/src/gateway.ts
```diff
@@ -156,6 +156,13 @@ export interface Gateway {
      */
     newSignedProposal(bytes: Uint8Array, signature: Uint8Array): Proposal;
 
+    /**
+     * Create a proposal from a serialized proposal.
+     * @param bytes - Serialized proposal.
+     * @returns A proposal.
```
bestbeforetoday, line 162: I wonder if we would be better to have a simple, consistent description for all of these `newType(bytes: Uint8Array)` methods. How about, "Recreate a _type_ from serialized data"?
```suggestion
     * Recreate a proposal from serialized data.
     * @param bytes - Serialized proposal.
     * @returns A proposal.
```

bestbeforetoday (changes requested):
One code naming change that needs to be reverted and doc wording suggestion, otherwise looks good

node/src/chaincodeeventsrequest.ts
sapthasurendran: It is already `SignedChaincodeEventsRequestOptions`

bestbeforetoday (approved):

PR #442, opened 2022-06-01
Java changes for deserialization 

sapthasurendran:
Contributes to #394 

### DIFF
--- a/java/src/main/java/org/hyperledger/fabric/client/GatewayImpl.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/GatewayImpl.java
@@ -148,15 +148,21 @@ final class GatewayImpl implements Gateway {
 
     @Override
     public Proposal newSignedProposal(final byte[] bytes, final byte[] signature) {
+        ProposalImpl result = newProposal(bytes);
+        result.setSignature(signature);
+        return result;
+    }
+
+    @Override
+    public ProposalImpl newProposal(final byte[] bytes) {
         try {
             ProposedTransaction proposedTransaction = ProposedTransaction.parseFrom(bytes);
             ProposalPackage.Proposal proposal = ProposalPackage.Proposal.parseFrom(proposedTransaction.getProposal().getProposalBytes());
             Common.Header header = Common.Header.parseFrom(proposal.getHeader());
             Common.ChannelHeader channelHeader = Common.ChannelHeader.parseFrom(header.getChannelHeader());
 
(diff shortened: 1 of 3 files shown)


java/src/main/java/org/hyperledger/fabric/client/GatewayImpl.java
```diff
@@ -148,14 +148,20 @@ public Network getNetwork(final String networkName) {
 
     @Override
     public Proposal newSignedProposal(final byte[] bytes, final byte[] signature) {
+        ProposalImpl result = (ProposalImpl) newProposal(bytes);
+        result.setSignature(signature);
+        return result;
+    }
+
+    @Override
+    public Proposal newProposal(final byte[] bytes) {
```
bestbeforetoday, line 157: Avoid the casts by returning the implementation type from the `newType()` methods. The client application works with the interface, which declares the interface type so this should be hidden from application code but avoids an unchecked cast.
```suggestion
        ProposalImpl result = newProposal(bytes);
        result.setSignature(signature);
        return result;
    }

    @Override
    public ProposalImpl newProposal(final byte[] bytes) {
```

java/src/main/java/org/hyperledger/fabric/client/GatewayImpl.java
```diff
@@ -148,14 +148,20 @@ public Network getNetwork(final String networkName) {
...
         try {
             ProposedTransaction proposedTransaction = ProposedTransaction.parseFrom(bytes);
             ProposalPackage.Proposal proposal = ProposalPackage.Proposal.parseFrom(proposedTransaction.getProposal().getProposalBytes());
             Common.Header header = Common.Header.parseFrom(proposal.getHeader());
             Common.ChannelHeader channelHeader = Common.ChannelHeader.parseFrom(header.getChannelHeader());
 
             ProposalImpl result = new ProposalImpl(client, signingIdentity, channelHeader.getChannelId(), proposedTransaction);
-            result.setSignature(signature);
             return result;
```
bestbeforetoday, line 165: If the `result` variable is only used as a return value in the following line, it is unnecessary. Just return the value directly:

```java
return new ProposalImpl(client, signingIdentity, channelHeader.getChannelId(), proposedTransaction);
```

bestbeforetoday (changes requested):
Two small changes (see comments), just to keep things tidy. They apply to all the `Gateway.newType()` methods.

bestbeforetoday (approved):

PR #445, opened 2022-06-07
Docs Cleanup

sapthasurendran:

### DIFF
--- a/java/src/main/java/org/hyperledger/fabric/client/Signable.java
+++ b/java/src/main/java/org/hyperledger/fabric/client/Signable.java
@@ -11,14 +11,15 @@ package org.hyperledger.fabric.client;
  */
 public interface Signable {
     /**
-     * Get the serialized proposal message bytes.
-     * @return A serialized proposal.
+     * Get the serialized message bytes.
+     * Serialized bytes can be used to recreate the object using methods on {@link Gateway}.
+     * @return A serialized signable object.
      */
     byte[] getBytes();
 
     /**
-     * Get the digest of the serialized proposal. This is used to generate a digital signature.
-     * @return A hash of the proposal.
+     * Get the digest of the signable object. This is used to generate a digital signature.
+     * @return A hash of the signable object.
      */
     byte[] getDigest();
 }
--- a/node/src/signable.d.ts
+++ b/node/src/signable.d.ts
@@ -9,12 +9,13 @@
  */
 export interface Signable {
     /**
-     * Get the serialized commit status request message.
+     * Get the serialized bytes of the signable object.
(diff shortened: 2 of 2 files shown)


java/src/main/java/org/hyperledger/fabric/client/Signable.java
```diff
@@ -11,14 +11,15 @@
  */
 public interface Signable {
     /**
-     * Get the serialized proposal message bytes.
-     * @return A serialized proposal.
+     * Get the serialized message bytes.
+     * Serialized bytes can be used to recreate the object using methods on Gateway.
```
bestbeforetoday, line 15: It would be good to use `{@link Gateway}` so users can more easily navigate to the related documentation.

node/src/signable.d.ts
```diff
@@ -9,12 +9,13 @@
  */
 export interface Signable {
     /**
-     * Get the serialized commit status request message.
+     * Get the serialized bytes of the signable object.
+     * Serialized bytes can be used to recreate the object using methods on Gateway.
```
bestbeforetoday, line 13: It would be good to use {@link Gateway} so users can more easily navigate to the related documentation.

bestbeforetoday (changes requested):
Just some convenience links would be nice -- see comments.

java/src/main/java/org/hyperledger/fabric/client/Signable.java
```diff
@@ -11,14 +11,15 @@
  */
 public interface Signable {
     /**
-     * Get the serialized proposal message bytes.
-     * @return A serialized proposal.
+     * Get the serialized message bytes.
+     * Serialized bytes can be used to recreate the object using methods on Gateway {@link Gateway}.
```
bestbeforetoday, line 15:
```suggestion
     * Serialized bytes can be used to recreate the object using methods on {@link Gateway}.
```

node/src/signable.d.ts
```diff
@@ -9,12 +9,13 @@
  */
 export interface Signable {
     /**
-     * Get the serialized commit status request message.
+     * Get the serialized bytes of the signable object.
+     * Serialized bytes can be used to recreate the object using methods on Gateway {@link Gateway}.
```
bestbeforetoday, line 13:
```suggestion
     * Serialized bytes can be used to recreate the object using methods on {@link Gateway}.
```

bestbeforetoday (changes requested):
You've duplicated the word **Gateway**. Check the generated API documentation to see how it looks. Suggested corrections below.

bestbeforetoday (approved):

PR #479, opened 2022-09-21
Update scenario test PEER_IMAGE_TAG to 3.0

denyeart:
Scenario tests utilize latest peer image pushed to hyperledger-fabric.jfrog.io/fabric-peer:amd64-latest. Currently this is version 3.0.
Update PEER_IMAGE_TAG to 3.0, this will also be used as the tag for the chaincode images (peer 3.0 will look for chaincode images 3.0).


bestbeforetoday (approved):
I think we really want to be testing against Fabric v2.5 development images, not v3.0 images, but merging as-is to get the build working again.

Issue #483, opened 2022-09-23
Support Node 18 LTS


bestbeforetoday:
**As a** TypeScript / JavaScript application developer
**I want** to use Node 18 LTS
**So that** I can remain current with the latest Node features and security fixes

Node 18 is scheduled to become the active LTS release on 2022-10-25:

- https://github.com/nodejs/release#release-schedule

Tasks:

- [x] Add Node 18 to the test matrix in the CI pipeline.
- [x] Add Node 18 to the compatibility matrix in the documentation (gh-pages branch).

vishwas031:
@bestbeforetoday I would like to work on this issue.

PR #487, opened 2022-10-07
support Node 18 LTS

vishwas031:
Signed-off-by: Vishwas Singh <031vishwas@gmail.com>

This PR changes the Node version to the latest version in the test matrix in the CI pipeline which should complete #483
@bestbeforetoday Sir, please review the PR.

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -37,7 +37,7 @@ variables:
   - name: GOVER
     value: 1.17
   - name: NODEVER
-    value: 18.x
+    value: 14.x
   - name: JAVAVER
     value: 8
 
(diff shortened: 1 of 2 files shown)


ci/azure-pipelines.yml
```diff
@@ -37,7 +37,7 @@ variables:
   - name: GOVER
     value: 1.17
   - name: NODEVER
-    value: 14.x
+    value: 18.x
```
bestbeforetoday, line 40: This one should remain at the minimum supported Node version (14.x) so the unit tests and packaging pick up accidental reliance on features that only exist in later Node versions

bestbeforetoday (changes requested):
Just one change to undo for this PR and it looks good.

A second PR to update the index.md compatibility matrix in the gh-pages branch will be needed to complete #483

ci/azure-pipelines.yml
vishwas031: I had made the changes. Please review it again.

PR #522, opened 2023-01-09
Evaluate,Submit, SubmitAsync with context

sapthasurendran:


pkg/client/contract.go
```diff
@@ -108,6 +133,42 @@ func (contract *Contract) Submit(transactionName string, options ...ProposalOpti
...
+	status, err := commit.StatusWithContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if !status.Successful {
+		return nil, fmt.Errorf("transaction %s rejected with validation code %d (%s)", status.TransactionID, int32(status.Code), peer.TxValidationCode_name[int32(status.Code)])
+	}
+
+	result := transaction.Result()
```
bestbeforetoday, line 168: The `SubmitAsyncWithContext()` already does most of the work and can be reused rather than duplicated.

The `Submit()` and `SubmitAsync()` implementations intentionally return the `result` along with any error. Depending on where the a occurs it may be possible to have a non-nil result and a non-nil error, and this is actually valid information the client might want to use. This function should behave the same way. If there isn't one already, there should probably be a unit test to specify this behaviour, although that would be my omission from the initial implementation so it doesn't need to be done in this PR.

```suggestion
	result, commit, err := contract.SubmitAsyncWithContext(ctx, transactionName, options...)
	if err != nil {
		return result, err
	}
	status, err := commit.StatusWithContext(ctx)
	if err != nil {
		return result, err
	}
	if !status.Successful {
		return nil, newCommitError(status.TransactionID, status.Code)
	}

```

bestbeforetoday (changes requested):
You've written a bit more code than necessary! See inline comments for where code can be reused to avoid duplication.

pkg/client/contract.go
```diff
@@ -108,6 +126,32 @@ func (contract *Contract) Submit(transactionName string, options ...ProposalOpti
...
+		return result, err
+	}
+
+	status, err := commit.StatusWithContext(ctx)
+	if err != nil {
+		return result, err
+	}
+
+	if !status.Successful {
+		return nil, fmt.Errorf("transaction %s rejected with validation code %d (%s)", status.TransactionID, int32(status.Code), peer.TxValidationCode_name[int32(status.Code)])
```
bestbeforetoday, line 149: This needs to return a CommitError, and there is an existing constructor function to create one.

```suggestion
		return nil, newCommitError(status.TransactionID, status.Code)
```

There should probably be a unit test to confirm that a CommitError is returned, like the one for Submit():

https://github.com/hyperledger/fabric-gateway/blob/68a5a1a875c874272fc3e280c61808f1145fdce0/pkg/client/submit_test.go#L159-L178

You might be able to convert some of the tests that are common between Submit() and SubmitWithContext() into parameterized tests. Whatever is easier.

bestbeforetoday (changes requested):

bestbeforetoday (approved):

andrew-coleman (approved):

Issue #614, opened 2023-08-01
End support for Node 16


bestbeforetoday:
**As a** project maintainer
**I want** to remove support for Node 16
**So that** features for currently supported Node runtimes can be exploited

Node 16 is scheduled to reach end-of-life on [2023-09-11](https://nodejs.org/en/blog/announcements/nodejs16-eol/). Once this date passes, support for Node 16 can be removed from the Fabric Gateway client API. However, this should probably be done after Node 20 becomes an LTS release on [2023-10-24](https://github.com/nodejs/release#release-schedule).

The following tasks are required:

- [ ] Remove Node 16 from GitHub Actions test matrix.
- [ ] Update tsconfig for node package and scenario tests to use [@tsconfig/node18](https://www.npmjs.com/package/@tsconfig/node18).
- [ ] Ensure the release minor version is updated since the last published release.

Issue #613, opened 2023-08-01
Support Node 20


bestbeforetoday:
**As a** Node application developer
**I want** to build and run my application with Node 20
**So that** I can stay current with latest features and bug fixes

[Node 20](https://github.com/nodejs/release#release-schedule) is scheduled to become an LTS release on 2023-10-24. Once available, the following tasks are required:

- [x] Add Node 20 to GitHub Actions test matrix
- [ ] Use Node 20 as the default Node version for version-agnostic GitHub Actions workflows
- [ ] Either on next release or if no code changes are required to pass tests, update documentation to include Node 20 in compatibility matrix.

PR #623, opened 2023-08-18
Removed Node 16 support and added Node 20

Shiva953:
Ref. #614 #613

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,7 +107,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 18
+          - 20
     env:
       SOFTHSM2_CONF: ${{ github.workspace }}/softhsm2.conf
     steps:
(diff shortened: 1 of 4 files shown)


bestbeforetoday:
Thank you for the contribution. As mentioned in #614, we cannot remove support for Node 16 until after it reaches end-of-life on [2023-09-11](https://nodejs.org/en/blog/announcements/nodejs16-eol/).

The Node implementation still needs to be built for the **minimum** supported Node release, not the latest. The changes to the `tsconfig.json` to build for a newer Node version will not work without also updating the devDependencies in `package.json`.

Pull request #621 is already open to include Node 20 in the test matrix. I am wary of using Node 20 for any of the other build steps until after it is actually promoted to active LTS status.

Since this PR contains only changes that cannot yet be merged, and changes that duplicate existing PRs, I am going to close this for now. Once Node 20 is promoted to an active LTS release, I would be happy for you to contribute the changes to remove Node 16 and update the TypeScript compile options to target Node 18. This would need to be applied to both the Node implementation and the Node scenario tests.

PR #625, opened 2023-09-03
Fixed Documentation Issue

rajat-dlt:


bestbeforetoday:
What issue do you see here? The actual numbers in Markdown lists (other than the first) are not relevant. This API documentation already renders correctly:

https://hyperledger.github.io/fabric-gateway/main/api/node/interfaces/Contract.html

rajat-dlt:
But still, this should be fixed. 
If one goes through the code, I guess.

bestbeforetoday:
> But still, this should be fixed. If one goes through the code, I guess.

Sorry I do not agree. TSDoc/TypeDoc aware IDEs (for example VS Code) render the numbering correctly in doc comments when mousing over the type. I prefer not to use explicit numbering (that will be ignored anyway) to avoid people having to remember to update the number sequence if adding or removing list entries, and the confusion it causes if they forget.

PR #657, opened 2023-11-14
chore(docs): fix nodejs async for loop examples 

jrasm91:
The block event examples do not seem to work and I have never heard of `for async`. I assume they are supposed to be `for await`, which is what the fabric samples use [here](https://github.com/hyperledger/fabric-samples/blob/c0a0104ca1ca9107cb7cea8af3758d1ec256df02/asset-transfer-events/application-gateway-typescript/src/app.ts#L79).


bestbeforetoday (approved):
Thank you for the correction!

PR #674, opened 2024-02-07
Vulnerability scan latest release commit

bestbeforetoday:


bestbeforetoday:
Closing since octokit/request-action used by the PR uses deprecated Node 16. Instead of scanning the release commit, continue to rely on a strategy of generally not updating runtime dependencies until a new release is being prepared to scan results reflect the status of the last release.

PR #682, opened 2024-02-29
Update remaining GitHub actions to avoid deprecated Node 16 versions

denyeart:
All three of them need to be updated!


bestbeforetoday (approved):
👍

PR #721, opened 2024-06-14
Use PMD and Spotless for Java code

bestbeforetoday:
Checkstyle only works with Java 11 and later, and does both code formatting and some static analysis. Replace Checkstyle with PMD for more thorough static analysis, and Spotless for formatting of Java code. Spotless also provides the capability to automatically reformat code to match the formatting rules.


bestbeforetoday:
For review it might be easier to look at the individual commits and ignore the one named _Apply Splotless formatting to Java code_, since it touches lots of files and is just a call to `mvn spotless:apply` to update formatting.

PR #730, opened 2024-07-02
Include generate Go mocks

bestbeforetoday:
Add generated Go mocks to source control. This makes it easier for new contributors to get started with the respository since no mocks need to be generated. It also avoids the need to generate mocks for CodeQL to be able to scan all Go source files.


bestbeforetoday:
@andrew-coleman you convinced me against this

PR #744, opened 2024-09-04
Use Ubuntu 24.04 for testing

bestbeforetoday:
Ubuntu 24.04.1 has now been released and is the current LTS release.


bestbeforetoday:
The switch to Ubuntu 22.04 caused the intermittent Java unit test failure to become very regular, which aided debugging. This change requires #746 to resolve the test issue before changing the Ubuntu version.

andrew-coleman (approved):

PR #756, opened 2024-09-26
Configure NPM to prefer dedupe in Node chaincode

bestbeforetoday:
When different versions of @grpc/grpc-js are resolved in the dependency tree, the following error in the chaincode can prevent the chaincode container from starting:

    TypeError: Channel credentials must be a ChannelCredentials object

This change adds an npmrc to the test Node chaincode, which configures NPM to perform dedupe by default. This ensures that only one version of @grpc/grpc-js is resolved in the chaincode's dependency tree.


bestbeforetoday:
incompatible version constraints fixed in fabric-chaincode-node.

PR #778, opened 2025-01-17
Seina/web impl

seina-webisoft:


bestbeforetoday:
Thank you for looking at this. It would be good to discuss what you've done and what you are trying to achieve. There are some significant problems with the approach taken here:

1. You are representing a huge amount of my work from PR #701 as your own in your commits.
2. You are including a gRPC implementation that is not used.
3. You have included an ECDSA signing implementation from the existing Node codebase that is not needed and was intentionally omitted since browsers contain their own Web crypto implementation.
4. You have added in digest and signing methods that are not needed and were intentionally omitted since message signing is done automatically using the client-supplied signing implementation.

Some confusion on the implementation above is understandable since the model is slightly different to the Node implementation, but the intended design and example code flows are available in the README.

seina-webisoft:
Hey @bestbeforetoday! Really sorry, I didn't mean to make a pull request here. I was experimenting off of your work and must have mistakenly created one.

seina-webisoft:
Regarding points 3 and 4, I reimplemented the digest and signing methods because I was having trouble signing transactions using the browser's SubtleCrypto API. I followed the README, but was receiving "unauthorized" errors when endorsing signed proposals. I can't say for certain why that would be the case, but when comparing the ECDSA signatures in both versions, I noticed the node version was deterministic whereas the browser one was not.

I also believe that, regardless of the existence of the browser's crypto implementation, it makes more sense to follow the same signing flow as the node library.

Finally, while the signing is done automatically in the client version, the method names (e.g., `newProposal`) match the unsigned variants in the node version. I think using more representative names, like `newSignedProposal`, can help avoid some confusion.

In any case, I apologize for the accidental PR haha. I'll be closing it now, but I would be glad to discuss these points further with you!

PR #796, opened 2025-04-21
Fix Node API documentation README

bestbeforetoday:
An update to TypeDoc resulted in the wrong README file being selected for the API documentation. This change explicitly specifies the README file to be used.

### DIFF
--- a/node/typedoc.json
+++ b/node/typedoc.json
@@ -1,8 +1,7 @@
 {
     "cleanOutputDir": true,
-    "entryPoints": [
-        "src/index.ts"
-    ],
+    "entryPoints": ["src/index.ts"],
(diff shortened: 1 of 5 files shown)


node/typedoc.json
```diff
@@ -1,8 +1,7 @@
 {
     "cleanOutputDir": true,
-    "entryPoints": [
-        "src/index.ts"
-    ],
+    "entryPoints": ["src/index.ts"],
     "out": "apidocs",
+    "readme": "src/README.md",
```
bestbeforetoday, line 5: This is the only functional change. The rest just updates to current dev dependency versions.

andrew-coleman (approved):

PR #866, opened 2025-07-17
Fix Node HSM signer failure

### LINKED ISSUE #865: Node HSM signer fails with @noble/curves@^1.9.2
The Node HSM signer fails with the following error when install resolves version 1.9.2 - 1.9.4 of the `@noble/curves` dependency:

> RangeError [ERR_INVALID_ARG_VALUE]: The argument 'size' is invalid. Received NaN

These versions introduce several new APIs, deprecate some existing APIs, and also introduce a regression where the length in octets of the base point order n (previously available from `p256.CURVE.nByteLength`) is `undefined`. The buffer allocated for the signature returned by the HSM is calculated as `nByteLength * 2`. With versions of `@noble/curves` including this regression, the result is `NaN`, which causes the error above.

bestbeforetoday:
A regression in the `@noble/curves` dependency causes the following error in the Node HSM signer if npm install resolves version 1.9.2 to 1.9.4.

This change uses new `@noble/curves` APIs (introduced in v1.9.2) and avoids newly deprecated API usage, including the one where the regression is introduced.

Closes #865

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -11,7 +11,7 @@
             "dependencies": {
                 "@grpc/grpc-js": "^1.12.0",
                 "@hyperledger/fabric-protos": "^0.3.0",
-                "@noble/curves": "^1.7.0",
+                "@noble/curves": "^1.9.4",
                 "google-protobuf": "^3.21.0"
             },
             "devDependencies": {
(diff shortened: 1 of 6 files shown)


bestbeforetoday:
@andrew-coleman This change is needed to fix the broken build in this repository and [fabric-samples](https://github.com/hyperledger/fabric-samples). Please can you review.

andrew-coleman (approved):

PR #1219, opened 2026-05-22
 bump protobufjs

MuthuSundaravadivel:
https://github.com/advisories/GHSA-jvwf-75h9-cwgg

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -8,6 +8,9 @@
             "name": "@hyperledger/fabric-gateway",
             "version": "1.11.1",
             "license": "Apache-2.0",
+            "overrides": {
+                "protobufjs": "^7.5.6"
+            },
             "dependencies": {
                 "@grpc/grpc-js": "^1.14.0",
                 "@hyperledger/fabric-protos": "^0.3.0",
--- a/node/package.json
+++ b/node/package.json
@@ -32,6 +32,9 @@
         "unit-test": "jest"
     },
     "license": "Apache-2.0",
+    "overrides": {
+        "protobufjs": "^7.5.6"
+    },
     "dependencies": {
         "@grpc/grpc-js": "^1.14.0",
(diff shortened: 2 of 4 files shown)


bestbeforetoday:
The resolved version of protobufjs was updated in commit bcea6d5dab73c5f96c7971a2ad9082068538086d to ensure a fixed version of protobufjs is used when running builds and tests. The override does not change the resolved dependency versions in package-lock.json since they already satisfy the override constraint.

For consumers of fabric-gateway, the version of protobufjs is dependent on the dependency resolution performed in their application. The version constraints of fabric-gateway allow fixed versions of protobufjs to be resolved.

MuthuSundaravadivel:
@bestbeforetoday 

The existing commit bcea6d5 updated protobufjs to 7.5.5 to address GHSA-xq3m-2v4x-88gg / CVE-2026-41242(https://github.com/advisories/GHSA-xq3m-2v4x-88gg). However, the newer advisory GHSA-jvwf-75h9-cwgg (https://github.com/advisories/GHSA-jvwf-75h9-cwgg)affects protobufjs versions prior to 7.5.6, so 7.5.5 is still considered vulnerable for this newer issue.

My intention with this PR was to ensure resolution to a non-vulnerable protobufjs version for the newer advisory. I understand your point that the override only affects dependency resolution within this repository and does not control downstream consumer resolution.

Please correct me if am wrong

bestbeforetoday:
The currently resolved version of protobufjs in the package-lock.json file is 7.6.1:

https://github.com/hyperledger/fabric-gateway/blob/032bdcbe20653d8c23c9fddb0293a7a55433f1c3/node/package-lock.json#L5297-L5299

https://github.com/hyperledger/fabric-gateway/blob/67d2824c5d13bfdf7f7155bbe3aa29baf03b70c1/scenario/node/package-lock.json#L2532-L2534

